PR #12, opened 2015-01-08
remove SystemUserSpawner.user_ids

minrk:
This should mean that the user_ids don't need to be stored ahead of time.

- remove user_ids dict
- only store single user_id on each Spawner
- load user_id from database state (allows Authenticator to set it)
- get default user_id from pwd, if not in db

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,16 +31,6 @@ you can use the SystemUserSpawner instead. Add the following to your
 
     c.JupyterHub.spawner_class = 'dockerspawner.SystemUserSpawner'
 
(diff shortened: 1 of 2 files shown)


minrk:
ping @jhamrick 

I'm testing this with the compmodels now, but so far it seems to work. Are there more use cases than

- local users (pwd.getpwname should work)
- Hub in docker on host (Authenticator can load user_id from restuser service, and store it in db state)

?

I guess subclasses can override how user_ids are looked up, but they still shouldn't need to build the dict beforehand.

PR #14, opened 2015-01-15
Add support for TLS authentication

jhamrick:
This makes it possible to connect to the docker daemon using TLS authentication, which is necessary (for example) if the docker container being launched is being created on a different machine.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -13,6 +13,7 @@ from jupyterhub.spawner import Spawner
 from IPython.utils.traitlets import (
     Dict,
     Unicode,
+    Bool
 )
 
 class DockerSpawner(Spawner):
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -69,6 +80,22 @@ def client(self):
         )
     )
 
+    tls = Bool(False, config=True, help="If True, connect to docker with --tls")
+    tls_verify = Bool(False, config=True, help="If True, connect to docker with --tlsverify")
+    tls_ca = Unicode("", config=True, help="Path to CA certificate for docker TLS")
+    tls_cert = Unicode("", config=True, help="Path to client certificate for docker TLS")
+    tls_key = Unicode("", config=True, help="Path to client key for docker TLS")
```
rgbkrk: Is there a way to specify the args that are bundled/required?

I'm currently seeing these option groupings:

* `tls=True` - standalone, causes `tls_config` to just be set to `True`.
* If `tls_cert` is set, `tls_key` needs to be as well.
* Any one of `tls_verify`, `tls_ca`, or the combination of `tls_cert` + `tls_key` need to be set for TLS.

jhamrick: Yeah, it probably would be good to document that. It corresponds essentially to this: https://docs.docker.com/articles/https/#client-modes , so the valid modes are:

* `tls=True`
* `tls_verify=True` and `tls_ca` is defined
* `tls_verify=True` and `tls_ca`, `tls_cert`, and `tls_key` are defined
* `tls_cert` and `tls_key` are defined

I think any other combination will cause docker-py to complain.

PR #15, opened 2015-01-18
DEV: Use exec for shell entrypoints.

ssanderson:
Makes the SIGTERM sent by `docker stop` hit the correct process.

### DIFF
--- a/singleuser/singleuser.sh
+++ b/singleuser/singleuser.sh
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-jupyterhub-singleuser \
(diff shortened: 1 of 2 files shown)


ssanderson:
cf https://github.com/ipython/docker-notebook/pull/37

ssanderson:
@minrk I don't think this actually works.  See https://github.com/ipython/docker-notebook/issues/38.  (I probably should have closed this).

minrk:
This has been reverted, it's no longer in master.

PR #16, opened 2015-02-07
Allow option to remove containers after they're stopped

jhamrick:
Adds the `remove_containers` config option, which is `False` by default, but if `True` will cause containers to be removed after they're stopped.


rgbkrk:
Your containers in compmodels all have volumes attached right?

We should make sure to remove volumes too (whatever is compatible with `docker rm -v <container>`).

jhamrick:
Good idea. I've updated the PR so that it removes volumes too.

PR #18, opened 2015-02-23
DEV: Barf more usefully on empty string for docker status.

ssanderson:
One of my containers died violently on startup and docker started returning an empty string for a status.  Not sure yet what causes this, but throwing an explicit error here is more useful than an `IndexError`, which was the previous behavior.


minrk:
:+1: always good to have more useful barf.

rgbkrk:
Thanks @ssanderson.

PR #20, opened 2015-02-24
DEV: Use `inspect_container` instead of parsing `containers`.

ssanderson:
Fixes https://github.com/jupyter/dockerspawner/issues/19

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -5,8 +5,10 @@ import itertools
 import os
 from textwrap import dedent
 from concurrent.futures import ThreadPoolExecutor
+from pprint import pformat
 
 import docker
+from docker.errors import APIError
 from tornado import gen
 
 from jupyterhub.spawner import Spawner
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -186,33 +188,41 @@ def poll(self):
...
+            self.container_id[:7],
+            pformat(container_state),
+        )
+
+        if container_state["Running"]:
             raise gen.Return(None)
-        elif parts[0] == 'Exited':
-            raise gen.Return(int(parts[1][1:-1]))
+        else:
+            raise gen.Return(
```
minrk: Since we require Python 3, we can just `return` instead of using `gen.Return`. There are some artifacts leftover from when I started the project, and was still considering support for Python 2.

ssanderson: Updated

dockerspawner/dockerspawner.py
```diff
@@ -185,34 +187,42 @@ def poll(self):
...
+            pformat(container_state),
+        )
+
+        if container_state["Running"]:
+            return None
+        else:
+            return (
+                "ExitCode={ExitCode}, "
+                "Error='{Error}', "
+                "FinishedAt={FinishedAt}".format(**container_state)
```
minrk: Is there any chance of KeyErrors in this format call? A previous bug could cause weird state when docker was having a hard time. Should we be less trusting of docker here? (AOK if the answer is no)

PR #21, opened 2015-02-28
Docker args config

bigzachattack:
Support both create_container and start optional args. Made args configurable, so you do not have to extend DockerSpawner to just change the args.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -87,8 +87,8 @@ class DockerSpawner(Spawner):
     tls_key = Unicode("", config=True, help="Path to client key for docker TLS")
 
     remove_containers = Bool(False, config=True, help="If True, delete containers after they are stopped.")
-    extra_create_kwargs = Dict({}, config=True, help="Additional args to pass for container create")
-    extra_start_kwargs = Dict({}, config=True, help="Additional args to pass for container start")
+    extra_create_kwargs = Dict(config=True, help="Additional args to pass for container create")
+    extra_start_kwargs = Dict(config=True, help="Additional args to pass for container start")
 
     @property
     def tls_client(self):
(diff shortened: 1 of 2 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -216,7 +218,7 @@ def get_container(self):
         self.container_id = ''
     
     @gen.coroutine
-    def start(self, image=None, **extra_create_kwargs):
+    def start(self, image=None):
```
jhamrick: I think we had talked about making it both an instance-level config option, and something that could still be passed in here by subclasses (e.g. in the case when you need per-container configuration, like [setting the working directory to be the home directory](https://github.com/compmodels/jupyterhub-deploy/blob/master/roles/jupyterhub_host/files/swarmspawner.py#L49)).

dockerspawner/dockerspawner.py
```diff
@@ -250,12 +252,17 @@ def start(self, image=None, **extra_create_kwargs):
...
+        # build the dictionary of keyword arguments for start
+        start_kwargs = dict(
             binds=self.volume_binds,
-            port_bindings={8888: (self.container_ip,)}
-        )
+            port_bindings={8888: (self.container_ip,)})
+        start_kwargs.update(self.extra_start_kwargs)
+
+        # start the container
+        resp = yield self.docker('start', self.container_id, **start_kwargs)
```
jhamrick: I don't think this `resp` is being used for anything?

bigzachattack:
@jhamrick Would ```def start(self, image=None, extra_create_kwargs={}, extra_start_kwargs={})``` work?

jhamrick:
Thanks! I would just say one other change -- could you make it be `extra_create_kwargs=None` and `extra_start_kwargs=None` in the function definition, rather than `{}`? There can be some nasty unpredictable bugs if you make default keyword objects be mutable objects. So if you set them to `None`, then you can check in the function body whether they're `None`, and if so, set them to `{}` there.

bigzachattack:
Thanks!

jhamrick:
This looks great! I actually just realized, though, that this change will also require updating `systemuserspawner.py` because it still assumes the `**kwargs` function signature: https://github.com/jupyter/dockerspawner/blob/master/dockerspawner/systemuserspawner.py#L119

Could you update that, too, so that it passes the working directory as part of `extra_create_kwargs`?

bigzachattack:
Any other place where this will trickle down that I need to go clean up?

jhamrick:
Perfect. I think those are all the places we need to worry about in this repository, at least.

PR #22, opened 2015-03-04
specify container ip on user.server

minrk:
I'm not 100% sure container_ip is always the right value for this, but I think it is.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -271,8 +271,9 @@ class DockerSpawner(Spawner):
         # start the container
         yield self.docker('start', self.container_id, **start_kwargs)
 
-        # get the public-facing port
+        # get the public-facing ip, port
         resp = yield self.docker('port', self.container_id, 8888)
+        self.user.server.ip = self.container_ip
         self.user.server.port = resp[0]['HostPort']
 
     @gen.coroutine


jhamrick:
I think this works, too.

In `SwarmSpawner` I apparently set `self.container_ip = "0.0.0.0"` and then set `self.user.server.ip` equal to the IP address of the machine that it was started on by swarm. I don't entirely remember why they're not the same, but I *think* it's just because `self.container_ip = "0.0.0.0"` is a useful catchall that allows the container to run publicly but `self.user.server.ip` needs to be the real IP so that the hub can communicate with the server. But, if the container IP were correctly set to the real IP, then that would still work I think -- the difficulty in that case is that the container IP isn't known until the container is started by swarm, but I'm handling that in any case, so yeah, I think it's fine.

bigzachattack:
I think 90% of the time setting self.container_ip is a right value for self.user.server.ip. The other 10% of the time, it would have to be handled in a subclass.

An option to handle more cases would be to have a `container_ip` (or `container_ip_bind`) and a `container_ip_connect`.   `container_ip`  would be used for binding the ports on the host side as is done today.  `self.user.server.ip` would be assigned `container_ip_connect` is specified, else it would be assiend `container_ip`.  Since `container_ip` defaults to `127.0.0.1` in the config, not specifying either should not change behavior.

With this option, DockerSpawner probably handles the cases outside of using Swam or specialized network/firewall configuration.

bigzachattack:
@minrk I implemented my idea at bigzachattack/dockerspawner@b72dc5276187e0e40484b877232f2d00dae52b39

I built it ontop of your commit and rebased from upstream.

PR #24, opened 2015-03-06
Allow use of usernames in volume paths

yuvipanda:
This allows us to give each docker instance a different 'home' 
somewhere on shared storage on the host, thus making it persistant.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -69,6 +69,8 @@ class DockerSpawner(Spawner):
             """
             Map from host file/directory to container file/directory.
             Volumes specified here will be read/write in the container.
+            If you use {username} in the host file / directory path, it will be
+            replaced with the current user's name.
             """
         )
     )
@@ -78,6 +80,8 @@ class DockerSpawner(Spawner):
             """
             Map from host file/directory to container file/directory.
             Volumes specified here will be read-only in the container.
+            If you use {username} in the host file / directory path, it will be
+            replaced with the current user's name.
             """
         )
     )
@@ -131,11 +135,11 @@ class DockerSpawner(Spawner):
         }
         """
         volumes = {
-            key: {'bind': value, 'ro': False}
+            key.format(username=self.user.name): {'bind': value, 'ro': False}
             for key, value in self.volumes.items()
         }
         ro_volumes = {
-            key: {'bind': value, 'ro': True}
+            key.format(username=self.user.name): {'bind': value, 'ro': True}
             for key, value in self.read_only_volumes.items()
         }
         volumes.update(ro_volumes)


yuvipanda:
Note that this assumes that the username is a valid file path component, but
note that the username is already used for the docker name, which is stricter
than file name components (no space, for example)

jhamrick:
This is pretty similar to what [SystemUserSpawner](https://github.com/jupyter/dockerspawner/blob/master/dockerspawner/systemuserspawner.py) does already -- is there some reason that doesn't work for you?

yuvipanda:
I'm using it with https://github.com/yuvipanda/mwoauthenticator which means
I do not have a system user corresponding to each JupyterHub user

jhamrick:
The name is a little misleading, the `SystemUserSpawner` doesn't actually require that the users exist on the host -- just that the user running inside the docker container is one that corresponds to the JupyterHub user (otherwise I think it's just always the `jupyter` user?). `SystemUserSpawner` essentially adds two things on top of `DockerSpawner`:

1. Ability to mount home directories from an arbitrary location on the host system
2. Creates the named user within the docker container with an appropriate UID

What's the ownership on the volumes you are mounting? I found that the UID of the volumes on the host needed to match the UID of the user running inside the docker container; otherwise the user didn't have permission to read/write their own home directory. That doesn't mean the user has to actually exist on the host system, but the UIDs still have to match.

yuvipanda:
(Note: I'm a Docker & IPython/Jupyter newbie...)

Right, so SystemUserSpawner looks like it does exactly what I would need, and would fix the permissions issues I'm having atm with this patch (the read-write volumes are mounted on /data/project/webhome/$username, and docker creates them when they are first needed, but for some reason those are created as belonging to 'root' and the docker processes seem to run as 'ubuntu'). 

However, SystemUserSpawner doesn't seem to work yet. I see that user_id doesn't have config set to True, so it'll be set from somewhere else. What is supposed to set them? The Authenticator? Without the user_id being set, it tries to use getent from pwd module and crashes...

jhamrick:
Don't worry, the rest of us are all still figuring stuff out as we go along too :-)

The way that the `user_id` gets set is from JupyterHub itself, which (somehow) knows about it. You can either hack the userlist or something to tell JupyterHub manually what all the UIDs are, or (in my case), the users actually do exist on some other system and JupyterHub goes and looks at that system to find out what the UIDs are (specifically, I'm using https://github.com/minrk/restuser to create and query users on the true host system).

However, this approach is not set in stone. @minrk and @rgbkrk and I actually just went to Docker today to talk about how we've been deploying stuff, and the Docker folks suggested that we have one UID for all the files, and have some other way of figuring out which home directories to mount. I think it would be worth investigating this option, and it would involve something like this:

* all the external volumes that we want to mount are chown'd to the same UID (e.g. 1000)
* JupyterHub knows the username because of the auth, and so knows which volume to actually mount (e.g. `/home/<user>`), though that volume is not chown'd to that user, but to e.g. UID 1000
* the volume is mounted inside the container as something like `/home/jovyan`, and the container is run as the `jovyan` user (which has UID 1000)

So basically in reality each user still gets their own homedir and persistent storage, but within their containers they all seem as if they have the same environment running as `jovyan`.

Does that setup sound like it would work for you?

yuvipanda:
Yup, that setup sounds ideal to me, esp. since I don't want to setup arbitrary users outside on the host system (we pull up users from LDAP, and anyone can register, there are bound to be clashes...)

Another problem would be that usernames aren't enforced as being filename / docker-name compatible, but that sounds like a auth plugin problem..

akhmerov:
Stepping into the discussion somewhat late. I like the setup which doesn't touch UID's on the host system somewhat more because it makes the setup easier to move/restore/cleanup/manage. That's why it seems this PR is quite useful. @RafalSkolasinski and I made a [deployment](https://github.com/RafalSkolasinski/volumespawner) very much like what @jhamrick outlined above, but this PR is much simpler, and seems to achieve the same aim.

minrk:
I think it's fine to allow usernames in paths, even with the rest of the discussion. I hadn't noticed just how long this PR has been open.

Issue #33, opened 2015-05-13
Passing certain arguments to start_container is all-but-deprecated


ssanderson:
Many arguments to start_container get translated by docker-py under the hood to entries in a host_config dictionary that's effectively deprecated in the latest docker.  From looking at https://github.com/docker/docker-py/pull/593/files, it looks like the following parameters are technically legal inputs to start_container, but are effectively deprecated:

```
start_config = utils.create_host_config(
             binds=binds, port_bindings=port_bindings, lxc_conf=lxc_conf,
             publish_all_ports=publish_all_ports, links=links, dns=dns,
             privileged=privileged, dns_search=dns_search, cap_add=cap_add,
             cap_drop=cap_drop, volumes_from=volumes_from, devices=devices,
             network_mode=network_mode, restart_policy=restart_policy,
             extra_hosts=extra_hosts, read_only=read_only, pid_mode=pid_mode,
             ipc_mode=ipc_mode, security_opt=security_opt, ulimits=ulimits
         )
 
         if isinstance(container, dict):
             container = container.get('Id')
 
         url = self._url("/containers/{0}/start".format(container))
         if not start_config:
             start_config = None
        elif utils.compare_version('1.15', self._version) > 0:
            warnings.warn(
                'Passing host config parameters in start() is deprecated. '
                'Please use host_config in create_container instead!',
                DeprecationWarning
            )
```

of these, I think the only ones we might be using are `binds` and `port_binds`, but `volumes_from`, `links`, and `extra_hosts` would be nice additions.

minrk:
closed by #34

PR #34, opened 2015-06-02
Use host_config in create_container

richafrank:
now that passing host_config parameters to start is deprecated.  (https://github.com/docker/docker-py/commit/329662c53ed13254449ef55742d2e8d577e0dd97)

For #33.

This also fixes usage of mem_limit and memswap_limit in our internal repo.

cc: @minrk @ssanderson

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -9,6 +9,7 @@ from pprint import pformat
 
 import docker
 from docker.errors import APIError
+from docker.utils import create_host_config
 from tornado import gen
 
 from jupyterhub.spawner import Spawner
@@ -281,6 +282,18 @@ class DockerSpawner(Spawner):
             if extra_create_kwargs:
                 create_kwargs.update(extra_create_kwargs)
 
+            # build the dictionary of keyword arguments for host_config
+            start_kwargs = dict(
+                binds=self.volume_binds,
+                port_bindings={8888: (self.container_ip,)})
+            start_kwargs.update(self.extra_start_kwargs)
+            if extra_start_kwargs:
+                start_kwargs.update(extra_start_kwargs)
+
+            host_config = create_host_config(**start_kwargs)
+            if host_config:
+                create_kwargs.setdefault('host_config', {}).update(host_config)
+
             # create the container
             resp = yield self.docker('create_container', **create_kwargs)
             self.container_id = resp['Id']
@@ -298,16 +311,8 @@ class DockerSpawner(Spawner):
             "Starting container '%s' (id: %s)",
             self.container_name, self.container_id[:7])
 
-        # build the dictionary of keyword arguments for start
-        start_kwargs = dict(
-            binds=self.volume_binds,
-            port_bindings={8888: (self.container_ip,)})
-        start_kwargs.update(self.extra_start_kwargs)
-        if extra_start_kwargs:
-            start_kwargs.update(extra_start_kwargs)
-
         # start the container
-        yield self.docker('start', self.container_id, **start_kwargs)
+        yield self.docker('start', self.container_id)
 
         # get the public-facing port
         resp = yield self.docker('port', self.container_id, 8888)
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -92,6 +92,7 @@ class DockerSpawner(Spawner):
     remove_containers = Bool(False, config=True, help="If True, delete containers after they are stopped.")
     extra_create_kwargs = Dict(config=True, help="Additional args to pass for container create")
     extra_start_kwargs = Dict(config=True, help="Additional args to pass for container start")
+    extra_host_config = Dict(config=True, help="Additional args to create_host_config for container create")
 
     hub_ip_connect = Unicode(
         "",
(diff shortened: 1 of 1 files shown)


ssanderson:
For reference, here are the arguments accepted by `Client.start` vs `create_host_config`:

```
In [15]: host_conf_args = set(inspect.getargspec(create_host_config).args)
In [16]: client_start_args = set(inspect.getargspec(Client.start.__wrapped__).args)
In [17]: host_conf_args
Out[17]:
{'binds',
 'cap_add',
 'cap_drop',
 'devices',
 'dns',
 'dns_search',
 'extra_hosts',
 'ipc_mode',
 'links',
 'log_config',
 'lxc_conf',
 'network_mode',
 'pid_mode',
 'port_bindings',
 'privileged',
 'publish_all_ports',
 'read_only',
 'restart_policy',
 'security_opt',
 'ulimits',
 'volumes_from'}

In [18]: client_start_args
Out[18]:
{'binds',
 'cap_add',
 'cap_drop',
 'container',
 'devices',
 'dns',
 'dns_search',
 'extra_hosts',
 'ipc_mode',
 'links',
 'lxc_conf',
 'network_mode',
 'pid_mode',
 'port_bindings',
 'privileged',
 'publish_all_ports',
 'read_only',
 'restart_policy',
 'security_opt',
 'self',
 'ulimits',
 'volumes_from'}

In [19]: host_conf_args - client_start_args
Out[19]: {'log_config'}

In [20]: client_start_args - host_conf_args
Out[20]: {'container', 'self'}
```

ssanderson:
@richafrank @minrk this makes sense to me, and it fixes a problem we were seeing with memory limits not being set correctly on the latest dockerspawner (@rgbrk I wouldn't be surprised if you've hit the same issue on tmpnb and friends if you're setting memory limits).  

One question is whether or not there should be some sort of deprecation of `extra_start_kwargs` in favor of something like `host_config_kwargs`.

ssanderson:
Also ping @rgbkrk, whose name I misspelled above.

minrk:
I think this makes sense in principle. I would probably make a new configurable `extra_host_config`, rather than load start_kwargs as a sub-keyword to create, which is a little confusing to me. The extra_start_kwargs helpstring can point to extra_host_config. If passing args to start doesn't make sense, we can deprecate it, but if other start_kwargs are likely to make sense, I'd leave it and point users to extra_host_config for the appropriate options.

ssanderson:
> If passing args to start doesn't make sense, we can deprecate it, but if other start_kwargs are likely to make sense, I'd leave it and point users to extra_host_config for the appropriate options.

Right now `Client.start` will accept the arguments listed above, and they're internally converted into a `HostConfig` dict that's sent to docker at container start time.  The issue with this is that passing many of these options at start time is deprecated, and my experience has been that they silently fail to have any effect.  In all cases where you'd want to pass something to `Client.start`, the correct thing to do (at least on recent versions of docker) is build a `HostConfig` dictionary, (either manually or via `docker.utils.create_host_config`) and pass that to `Client.create_container`, which is what's being done under the hood by this PR.  It sounds like the counterproposal you're making is that we should just deprecate `extra_start_kwargs` entirely in favor of something like `host_config_kwargs`?  I think that's reasonable as long as we're ok with breaking existing code that's using `extra_start_kwargs`.  I'm not sure who that would affect beyond @jhamrick.

rgbkrk:
This whole time I didn't realize you could pass ulimit options in through docker-py.

rgbkrk:
>  I think that's reasonable as long as we're ok with breaking existing code that's using extra_start_kwargs. I'm not sure who that would affect beyond @jhamrick.

Care to post to the mailing list? I think people are just getting started with jupyterhub and haven't made nearly as complex deployments as you @ssanderson / @quantopian or @jhamrick, but I could be wrong. The compmodels deployment is done for the semester, so it wouldn't be bad per se.

ssanderson:
Jupyter group post here: https://groups.google.com/forum/#!topic/jupyter/n9wyBMn2YuQ

bigzachattack:
I am with @minrk on providing a new configurable `extra_host_config`.  That would be an easy change for anyone using `extra_start_kwargs`.

I do have a JupyterHub deployment that uses dockerspawner.  At the moment, I am using `extra_start_kwargs` in my config file to enable privileged containers.

richafrank:
I can push a version that adds a new ```extra_host_config``` and keeps the existing ```extra_start_kwargs``` for backwards compatibility.  It should remove any confusion around ```start_kwargs``` actually being passed to ```create_container``` instead of ```start```, as the current version does.

richafrank:
Updated as described with https://github.com/quantopian/dockerspawner/commit/168ecabaa72508906e21f3971b9f34d547ba1f38

richafrank:
Thanks for reviewing, @rgbkrk .  Now that it's backwards compatible, anything else to do here?  (Should I squash?)

rgbkrk:
I waited, just in case @jhamrick had something to chime in on.

richafrank:
Ahh, cool, sounds good.

minrk:
:+1:

jhamrick:
Hi all, sorry for being unresponsive on this, have had a busy week with a paper deadline... in any case I'm fine with this change -- I think the only thing I was using was `extra_create_kwargs` rather than `extra_start_kwargs`.

minrk:
@jhamrick thanks!

minrk:
Thanks @richafrank and @ssanderson for putting this together.

richafrank:
Thanks all!

rgbkrk:
Thank you!

PR #39, opened 2015-07-17
"Sanitize" docker container name by using hex representation of the user name

martijnthe:
Only `[a-zA-Z0-9][a-zA-Z0-9_.-]` are allowed in the container name.
I ran into this issue when using Google OAuth, where the user name is the full email address.
Converting the user name to hex was the simplest, fool-proof thing I could think of.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -1,6 +1,7 @@
 """
 A Spawner for JupyterHub that runs each user's server in a separate docker container
 """
+import codecs
 import itertools
 import os
 from textwrap import dedent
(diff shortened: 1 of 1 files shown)


minrk:
I think we can use escaping to avoid garbled names in common cases.

martijnthe:
Hmm, I doubt it.

FYI, I see this person also had this issue:
https://github.com/jupyter/dockerspawner/issues/25

minrk:
@martijnthe you doubt that escaping works?

rgbkrk:
Not sure how we'd escape an email address with the specific allowance of container names.

I don't want to see container names becoming hexadecimal, though it may be necessary if there isn't a way to sanitize it in the auth layer.

minrk:
I don't think escaping is complicated. We can make a strict safe character list and escape everything else.

minrk:
I took a stab at escaping container names in #40.

martijnthe:
@minrk that should work :+1:

PR #40, opened 2015-07-17
escape username in container name

### LINKED ISSUE #25: Support usernames that do not conform to docker container names
Some authentication providers (cough mediawiki) allow a lot more characters classes than they should (mediawiki allows (, spaces, ), etc). Hashing or seperating 'userid' from 'username' would allow people with such special characters in their username to use jupyterhub.

minrk:
only allow `[A-Za-z0-9_-]`, escape everything else.

cf [docker container name spec](https://github.com/ajhager/docker/commit/f63cdf0260cf6287d28a589a79d3f947def6a569)

closes #25
closes #39

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -161,14 +161,19 @@ class DockerSpawner(Spawner):
         volumes.update(ro_volumes)
         return volumes
 
+    _escaped_name = None
     @property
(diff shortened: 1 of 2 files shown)


rgbkrk:
Ah, so it's a prescribed format for escaping through a library. Reasonable.

PR #41, opened 2015-07-27
Add a use_docker_client_env option

alexstrat:
I was not able able to make dockerspawner work with boot2docker without hardcoding dockerspawner code.

Here is an added option that tells `DockerSpawner` to use [`docker.utils.kwargs_from_env`](https://docker-py.readthedocs.org/en/latest/boot2docker/) to build the docker client.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -10,7 +10,7 @@ from pprint import pformat
 
 import docker
 from docker.errors import APIError
-from docker.utils import create_host_config
+from docker.utils import create_host_config, kwargs_from_env
 from tornado import gen
 
 from escapism import escape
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -38,18 +38,24 @@ def client(self):
...
-            if self.tls:
-                tls_config = True
-            elif self.tls_verify or self.tls_ca or self.tls_client:
-                tls_config = docker.tls.TLSConfig(
-                    client_cert=self.tls_client,
-                    ca_cert=self.tls_ca,
-                    verify=self.tls_verify)
+            if self.use_docker_client_env:
+                kwargs = kwargs_from_env()
+                kwargs['tls'].assert_hostname = False
```
minrk: Is it appropriate to hardcode this? It doesn't seem so.

alexstrat: Yes indeed! Fixed in 50ef29d43ac35f3abca64193b0122f5f54e93593

rgbkrk: Thanks, we definitely use TLS properly on our swarm deployment.

alexstrat:
@minrk do you consider it ready to be merged?

rgbkrk:
Looking good to me. If you're able to rebase, I'd rather not have the merge commit.

rgbkrk:
Thanks again @alexstrat!

PR #43, opened 2015-08-05
Add option to toggle hostname assertion with TLS

jhamrick:
When using TLS, sometimes the certs don't exactly match the hostname where the docker daemon is running. For example, the cert might be for `*` but docker-py will complain if `assert_hostname` is true (which is the default) because `*` is not the same as the actual hostname.

I don't think the docker command line client actually does this assertion, because accessing the docker daemon via the command line with TLS works fine, it's just through docker-py that it doesn't by default (but it does when the assertion is disabled).


minrk:
Makes sense, thanks.

PR #44, opened 2015-09-21
add version=auto to docker client to avoid verison conflict

jlamcanopy:


minrk:
This makes sense to me. cc @jhamrick @rgbkrk @ssanderson for more Docker expertise.

rgbkrk:
Yup, we just switched over tmpnb to this as well not too long ago. I thought it defaulted to `version=auto` [in docker-py though](https://github.com/docker/docker-py/blob/e7ad536168a1dfa3f85b124411055fde0400b74f/docker/client.py#L71-L76), so I think that's what led me to not adding it on here. As linked though, that's not the case.

rgbkrk:
As for what this does, docker-py does a version call to the server and sets its API version according to the server's response. I'm :+1: on adding this to dockerspawner.

ssanderson:
cc @richafrank and @llllllllll.  I think this might fix some of the issues you guys ran into a while back when trying to upgrade one of docker/docker-py but not the other?

richafrank:
We need this functionality in the `use_docker_client_env` case as well.  I just opened a PR that includes that: https://github.com/jupyter/dockerspawner/pull/47

PR #49, opened 2015-10-13
MAINT: tls_assert_hostname should be either a string or False

richafrank:
according to [urllib3] (https://urllib3.readthedocs.org/en/latest/pools.html#urllib3.connectionpool.HTTPSConnectionPool):  "VerifiedHTTPSConnection uses one of assert_fingerprint, assert_hostname and host in this order to verify connections. If assert_hostname is False, no verification is done."

With the current default value of `True`, I get an AttributeError when urllib3 tries to call `lower` on `True` in order to compare it to another string:
```
      File "/Users/rich/.virtualenvs/research_env/lib/python3.4/site-packages/requests/packages/urllib3/connection.py", line 253, in connect
        match_hostname(cert, self.assert_hostname or hostname)
      File "/usr/local/Cellar/python3/3.4.3_2/Frameworks/Python.framework/Versions/3.4/lib/python3.4/ssl.py", line 268, in match_hostname
        if _dnsname_match(value, hostname):
      File "/usr/local/Cellar/python3/3.4.3_2/Frameworks/Python.framework/Versions/3.4/lib/python3.4/ssl.py", line 225, in _dnsname_match
        return dn.lower() == hostname.lower()
    AttributeError: 'bool' object has no attribute 'lower'
```

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -18,8 +18,11 @@ from jupyterhub.spawner import Spawner
 from traitlets import (
     Dict,
     Unicode,
-    Bool
+    Bool,
+    Union,
+    Enum,
 )
+
 
 class DockerSpawner(Spawner):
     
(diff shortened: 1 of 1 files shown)


richafrank:
@rgbkrk Any thoughts on this one?

rgbkrk:
You've been moving this right along with dockerspawner, thanks so much!

I'm going to give you commit access to help us maintain dockerspawner, since you're so active and I trust in what you code and test here. PRs as always, never straight to master.

richafrank:
Thanks!  Glad to help - sounds good!

rgbkrk:
Same with @ssanderson. Thanks all.

PR #50, opened 2015-10-15
fix traitlets usage on tls_assert_hostname

minrk:
- config=True, help at the top-level
- add None to Enum

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -19,9 +19,16 @@ from traitlets import (
     Dict,
     Unicode,
     Bool,
-    Union,
     Enum,
 )
+
+
+class UnicodeOrFalse(Unicode):
+    info_text = 'a unicode string or False'
+    def validate(self, obj, value):
+        if value is False:
+            return value
+        return super(UnicodeOrFalse, self).validate(obj, value)
 
 
 class DockerSpawner(Spawner):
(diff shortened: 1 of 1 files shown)


minrk:
Final touches on #49

richafrank:
@minrk It looked like metadata on Union [doesn't actually work] (https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L1471), so I moved it down to the [first element] (https://github.com/ipython/traitlets/blob/master/traitlets/traitlets.py#L1499).  Did you find that the traitlet still got configured?

minrk:
@richafrank weird, I did test it and it worked, but I'll dig in again and be more rigorous this time. I suspect that's an inaccurate docstring, but it may be a difference between stable and master of traitlets.

minrk:
Yup, after testing confirmed that this works on both stable and master traitlets. I'll go look into the Union docstring to see what's up there.

richafrank:
With this branch, `config=True` is working for me; however, I'm getting `traitlets.traitlets.TraitError: No default value found for tls_assert_hostname trait`, which makes sense, since no default is defined.  If I add `default_value=None` to the Union, then the default works, but `config=True` no longer works.  

Looks like if a default is defined for the Union, then it will get validated and reset the metadata to that of the trait_type that it matches.

I recognize one doesn't need both cases at the same time, but I was hoping to support either case.  I *think* #49 did that, but it's not the cleanest code because it's working around this odd behavior.  

Are you seeing this too?  I'm on traitlets==4.0.0.

minrk:
Arg, Union is a bit of a mess, I've never used it before and after this, I probably never will. I'll do a different fix with a much simpler custom traitlet.

minrk:
Done without Union now. I wouldn't recommend using Union to anyone, really.

richafrank:
> I wouldn't recommend using Union to anyone, really.

I concur.

The custom traitlet looks good to me - I'll make the same change to https://github.com/quantopian/DockORM/pull/3.

minrk:
Thanks!

PR #57, opened 2015-11-30
Allow usage of internal docker subnet IP

andreaslang:
The context for this proposed change is, that I want to be able to use internal docker IP addresses instead of the public ones. I am running the jupyterhub server (as docker container) and the user containers on the same docker engine which is enough for smaller scenarios without a swarm. If I use the code as it is the public ip/port  of the spawned user container is used (localhost and random unprivileged port). These changes add the possibility to use the internal IP instead and (since it is not required in that case) do not expose a port to the outside.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -135,6 +135,18 @@ class DockerSpawner(Spawner):
         )
     )
 
+    use_internal_ip = Bool(
+        False,
+        config=True,
+        help=dedent(
+            """
+            Enable the usage of the internal docker ip. This is useful if you are running
+            jupyterhub (as a container) and the user containers within the same docker engine.
+            E.g. by mounting the docker socket of the host into the jupyterhub container.
+            """
+        )
+    )
+
     @property
     def tls_client(self):
         """A tuple consisting of the TLS client certificate and key if they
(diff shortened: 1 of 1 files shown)


minrk:
+1 from me. @ssanderson, @jhamrick?

andreaslang:
I realised one little issue with my code. It only works if the docker containers are in the default network (which is called bridge), otherwise _NetworkSettings.IPAddress_ is not set. This happens if you add a new network with _docker network create_ and add _network\_mode=\<network_name\>_ to _extra\_start\_kwargs_. I would propose fixing that in a second pull request after this one is merged (assuming it will be merged).

I think the cleanest way of dealing with that is adding another config option specifying the network the internal ip address should be taken from (defaulting to bridge). Then the _NetworkSettings.Networks.\<name\>.IPAddress_ property can be used.

Let me know if you prefer pushing the necessary changes here or in another pull request.

jhamrick:
:+1: from me (though I haven't tested this manually).

It sounds fine for the other changes regarding the network to go in a separate pull request. Thanks!

dockerspawner/dockerspawner.py
```diff
@@ -360,10 +372,20 @@ def start(self, image=None, extra_create_kwargs=None,
...
-        self.user.server.ip = self.container_ip
-        self.user.server.port = resp[0]['HostPort']
+        ip, port = yield from self.get_ip_and_port()
+        self.user.server.ip = ip
+        self.user.server.port = port
+
+    def get_ip_and_port(self):
+        if self.use_internal_ip:
+            inspect_resp = yield self.docker('inspect_container', self.container_id)
+            ip = inspect_resp['NetworkSettings']['IPAddress']
```
ssanderson: Does this need to be done every time we spawn?  It might be nice to make `internal_ip` another traitlet with a default value that's dynamically populated once the first time we need it (which would only happen if `use_internal_ip` were true.

andreaslang: Should be possible, but just to be sure, as I don't know the code very well: Does every user get its own spawner? As those are the IPs for the user servers/containers so every user gets a different IP.

minrk: Yes, there's a Spawner object per single-user server.

andreaslang: OK, then I just need to test if docker re-assigns internal IPs between restarts of containers. If not I'll implement and push the suggested change later today.

andreaslang: Not sure if you get notified if I edit my original post so here as separate post:

Tests done. unfortunately, even if the container is not removed but just stopped, it can happen that the IP address is reassigned. This means if the user manually stops containers or removes them the IP address might be outdated, so I don't think caching is an option here. Let me know how you want to proceed.

andreaslang:
Hi just wanted to follow up on this. As mentioned in my earlier comment, Docker can reassign internal IPs which is why the IP address should be loaded on every spawn. Using a dynamically populated traitlet could cause errors. Is there anything else to be done on this pull request?

minrk:
Nope, I think we are good. Thanks, @andreaslang!

PR #58, opened 2015-12-21
Allow usage of custom docker networks with internal ip

andreaslang:
As announced in my UseInternalIP pull request, this is the change necessary for docker 1.9 networks. It allows you to have dockerized jupyterhub and the containers in a non-standard docker network. In my use case I want to restrict the access of jupyterhub containers (non-jupyterhub related) to other docker containers in the same deployment. Just a simple security measure.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -152,8 +152,8 @@ class DockerSpawner(Spawner):
         config=True,
         help=dedent(
             """
-            Retrieve the internal IP from this docker network. Defaults to the default docker
-            network 'bridge'. You need to set this if you run your jupyterhub containers in a
+            The name of the docker network from which to retrieve the internal IP address. Defaults to the default
+            docker network 'bridge'. You need to set this if you run your jupyterhub containers in a
             non-standard network. Only has an effect if use_internal_ip=True.
             """
         )
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -391,22 +391,30 @@ class DockerSpawner(Spawner):
     def get_ip_and_port(self):
         if self.use_internal_ip:
             resp = yield self.docker('inspect_container', self.container_id)
-            networks = resp['NetworkSettings']['Networks']
-            if self.internal_ip_network_name not in networks:
-                raise Exception(
-                    "Unknown docker network '{network}'. Did you create it with 'docker network create <name>' and "
-                    "did you pass network_mode=<name> in extra_kwargs?".format(
-                        network=self.internal_ip_network_name
-                    )
-                )
-            network = networks[self.internal_ip_network_name]
-            ip = network['IPAddress']
+            network_settings = resp['NetworkSettings']
+            if 'Networks' in network_settings:
+                ip = self.get_network_ip(network_settings)
+            else:  # Fallback for old versions of docker (<1.9) without network management
+                ip = network_settings['IPAddress']
             port = 8888
         else:
             resp = yield self.docker('port', self.container_id, 8888)
             ip = self.container_ip
             port = resp[0]['HostPort']
         return ip, port
+
+    def get_network_ip(self, network_settings):
+        networks = network_settings['Networks']
+        if self.internal_ip_network_name not in networks:
+            raise Exception(
+                "Unknown docker network '{network}'. Did you create it with 'docker network create <name>' and "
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -147,6 +147,18 @@ def client(self):
         )
     )
 
+    internal_ip_network_name = Unicode(
+        "bridge",
+        config=True,
+        help=dedent(
+            """
+            Retrieve the internal IP from this docker network. Defaults to the default docker
```
minrk: Should this be `network` or `network_name`? Also, the first line should probably start with:

> The name of the docker network from which to retrieve the internal IP address.

From this first line and the variable name, I expected this trait to be an IP address, not the name of the network.

minrk:
cc @jhamrick @rgbkrk @ssanderson for docker review

dockerspawner/dockerspawner.py
andreaslang: It is the name. For example created with _docker network create <network-name>_. I did put the internal_ip in front of it as it only makes sense to use this setting if you use _use_internal_ip=True_. Would you prefer something like just _network_name_ or _internal_network_name_?

minrk: `network_name` or `internal_network_name` seems clearer.

minrk: Also worth adding a line to the help:

> Only has an effect if `use_internal_ip=True`.

andreaslang: done

minrk: Can you make the mentioned update to the help string? "The name of the network..."

dockerspawner/dockerspawner.py
```diff
@@ -147,6 +147,18 @@ def client(self):
...
+        "bridge",
+        config=True,
+        help=dedent(
+            """
+            The name of the docker network from which to retrieve the internal IP address. Defaults to the default
+            docker network 'bridge'. You need to set this if you run your jupyterhub containers in a
+            non-standard network. Only has an effect if use_internal_ip=True.
+            """
+        )
+    )
```
rgbkrk: A+, really happy to see someone add networks support here.

dockerspawner/dockerspawner.py
```diff
@@ -378,8 +390,12 @@ def start(self, image=None, extra_create_kwargs=None,
...
     def get_ip_and_port(self):
         if self.use_internal_ip:
-            inspect_resp = yield self.docker('inspect_container', self.container_id)
-            ip = inspect_resp['NetworkSettings']['IPAddress']
+            resp = yield self.docker('inspect_container', self.container_id)
+            network_settings = resp['NetworkSettings']
+            if 'Networks' in network_settings:
+                ip = self.get_network_ip(network_settings)
+            else:  # Fallback for old versions of docker (<1.9) without network management
+                ip = network_settings['IPAddress']
```
rgbkrk: Sweet.

dockerspawner/dockerspawner.py
```diff
@@ -387,6 +403,19 @@ def get_ip_and_port(self):
             port = resp[0]['HostPort']
         return ip, port
 
+    def get_network_ip(self, network_settings):
+        networks = network_settings['Networks']
+        if self.network_name not in networks:
+            raise Exception(
+                "Unknown docker network '{network}'. Did you create it with 'docker network create <name>' and "
+                "did you pass network_mode=<name> in extra_kwargs?".format(
```
rgbkrk: Wonderful error message!

rgbkrk:
Thanks for this PR @andreaslang!

One possibility I was going to use docker networks for is creating isolated networks (for anonymous users):

```bash
docker network create -o com.docker.network.bridge.enable_icc=false \
                      -o com.docker.network.bridge.enable_ip_masquerade=false \
                      isolated
```

Though I guess I'm not sure how I'd connect back to the hub with that... Maybe only useful with tmpnb. :-/

andreaslang:
@rgbkrk what I am doing is having an adjusted jupyterhub startup script which makes sure my spawn network exists, creates it if it doesn't and connects the jupyterhub docker container to the network. As containers can be in more than one network this works even if you want jupyterhub in another network. The rest is some jupyterhub configuration magic to get the right IPs. I am planning to create a simplified example based on the setup we are using right now and to put that into a pull request.

minrk:
Thanks, @andreaslang!

PR #69, opened 2016-02-08
Added Docker container linking

Fokko:
Added Docker container linking so that the user-container gets linked to the jupyterhub container, which makes communication much easier.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -31,7 +31,7 @@ class UnicodeOrFalse(Unicode):
 
 
 class DockerSpawner(Spawner):
-    
+
     _executor = None
     @property
     def executor(self):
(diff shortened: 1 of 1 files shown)


rgbkrk:
This is fine with me, though I'll never use it as linking is on the table for deprecation in Docker (relying on docker networks and volumes).

Fokko:
@rgbkrk thanks. I would like to see this merged into jupyterhub and upgrade to networks in the short future.

I've pushed a small update. I also link to a Spark container, but had some problems of overwriting keys in the dicts as `.update` does not work recursively.

dockerspawner/dockerspawner.py
```diff
@@ -150,6 +150,18 @@ def client(self):
         )
     )
 
+    use_docker_links = Bool(
```
minrk: Shouldn't `links` be a dict, rather than a bool for hardcoded values?

Fokko: Definitely.

rgbkrk:
Whoa, just noticed we're using PyFlakes on a Jupyter project. That's awesome.

minrk:
@rgbkrk I added it as a band-aid while there are zero real tests.

PR #72, opened 2016-02-16
Use HostIp from docker port output

smashwilson:
Rather than use `container_ip` from the DockerSpawner configuration, use the IP address reported by the `docker port` command. This should behave identically when `container_ip` is left at its default or specified explicitly, because Docker will report the IP that was bound. But, by using `container_ip` as the interface to bind and reading the read address from Docker after container launch, users can specify:

```python
c.DockerSpawner.container_ip = "0.0.0.0"
```

And JupyterHub will connect to the launched container at its public IP.

With only this change, I was able to use the regular `DockerSpawner` to run JupyterHub on Swarm, with user containers launching on different nodes. :confetti_ball:

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -422,7 +422,7 @@ class DockerSpawner(Spawner):
             port = 8888
         else:
             resp = yield self.docker('port', self.container_id, 8888)
-            ip = self.container_ip
+            ip = resp[0]['HostIp']
             port = resp[0]['HostPort']
         return ip, port
 


minrk:
Thanks!

rgbkrk:
:bow:

rgbkrk:
Verified that this works on a fresh deployment:

```
$ docker ps
CONTAINER ID        IMAGE                COMMAND                  CREATED             STATUS              PORTS                                                      NAMES
d493933f6257        jupyter/singleuser   "tini -- sh /srv/sing"   11 minutes ago      Up 11 minutes       146.20.68.20:32768->8888/tcp                               c0fa2c6f-2045-4ca0-85e6-2d3addd0e135-n2/jupyter-rgbkrk
803f72a69ce9        jupyterhub-carina    "jupyterhub -f /srv/j"   12 minutes ago      Up 11 minutes       172.99.73.183:8081->8081/tcp, 172.99.73.183:80->8000/tcp   c0fa2c6f-2045-4ca0-85e6-2d3addd0e135-n1/jupyterhub
```

:tada:

PR #73, opened 2016-02-16
[WIP] Return None even if container does not exist

betatim:
This follows on from jupyter/jupyterhub#428. There are cases when the container is still being created by `start()` and `poll()` is being called, in which case `jupyterhub` decides that things went wrong.

As a starting point I suggest to return `None` even if the container doesn't exist. This is probably not the best solution though as the container might not exist because it got killed. How to best handle the difference between "container has never existed so might still be being created" and "container used to exist, but not anymore -> exited"?

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -301,7 +301,7 @@ class DockerSpawner(Spawner):
         container = yield self.get_container()
         if not container:
             self.log.warn("container not found")
-            return ""
+            return None 
 
         container_state = container['State']
         self.log.debug(


rgbkrk:
`None` seems more appropriate than `""` here, so :+1:.

Later on in the same function if `container_state["Running"]` is `True`, this returns `None`. Do these semantics match well or do we need to report something more robust?

betatim:
The semantics are "return `None` if not yet exited" so I think that is fine.

The tricky part is detecting and returning an integer:
 1. that the container has been built successfully, but now doesn't exist anymore
 2. building the container has failed

Actually, I wonder if (2) should still return `None` as the spawner didn't exit (because it never started ...)

minrk:
@betatim failing to build the container should result in non-None poll result.

It seems like the main issue to address is an async race condition where start may yield before the container has been created (in fact, it yields on the call to create the container). The simplest fix is probably to set a flag in start that's cleared as soon as the container has (or has not) been created. Something like:

```python
def start():
    ...
    self._no_container_yet = True
    try:
        container_id = yield docker.container...
    finally:
        self._no_container_yet = False
...
def poll():
    if self._no_container_yet:
        return None
    ...
```

carolynvs:
I ran into this last week when trying to spawn containers using an image that had not already been pulled down to the swarm.

In case it helps here's what I did to get it working (e.g. the UI will detect that I'm a slow starter and show a "Spinning up your server" wait page).

1. Don't return from start until the container is completed. I did this by first pulling in start, waiting for the pull to complete, then letting DockerSpawner create the container https://github.com/carolynvs/howtowhale/blob/master/howtowhale-hub/lib/jupyterhub_carina/CarinaSpawner.py#L65.
2. I ended up needing both a Started and Starting flag because `poll()` is called before `start()`, and then again before `start()` is complete (after it first yields). Without both variables, I was being dumped to home screen without `start()` ever being called, and when I clicked the  "Start Server" button, I'd get stuck in an infinite redirect loop. https://github.com/carolynvs/howtowhale/blob/master/howtowhale-hub/lib/jupyterhub_carina/CarinaSpawner.py#L93.

I also tried (and removed) having `poll()` return `None` instead of `""` when the container wasn't started but that didn't help all on its own. It seemed at the time that sometimes `poll()` should return `None` and other times it shouldn't and the rules were different if I was called before I had been started or after.

It probably doesn't need to be this complicated but /shrug, this is the glue I needed to get it working.

betatim:
I will look at your code @carolynvs. What @minrk suggested is close to what I did for `everware` but wanted to hear some feedback. I will update the PR later today.

betatim:
The problem with setting flags is that I can't work out how this would generalise/work in the case were people are inheriting from `DockerSpawner` and doing things in their `start` that cause the race. Insisting that every overload contains the same few lines to set these flags is ugly. Is there something in the python box of tricks to do that for us? (channel your inner David Beazley)

minrk:
I'm not sure why subclasses would need to be copying anything. There should only be a race on the `yield make_container()`.

I think we can make this simpler upstream, without needing any flags or anything in DockerSpawner.
I think I can guarantee that the Hub will never call `Spawner.poll` before `Spawner.start` has returned. Then you don't need to worry about this. In that case, I think the returncode should be 0 (or -1, it doesn't really make a difference) when there's no container, since the server really isn't running.

rgbkrk:
With #442, how's this now @betatim?

carolynvs:
Ooh, that bug fix is mighty helpful. I can give it a try tonight.

betatim:
LGTM :rocket: closing this

carolynvs:
I finally got a chance to try #442 out and it is not working as I had hoped. Here is what I'm seeing:

1. poll is called, no container is found.
2. start is called, which kicks off a very slow command (pulling a 4GB docker image) then yields
3. poll is called before start is finished.

I was hoping that poll wouldn't be called until start was finished? Is there a timeout somewhere that I need to configure to buy more time?

carolynvs:
Sorry for the deleted comment! I've tried out the latest jupyterhub docker image and my polling problems are fixed. Thank you! :cake:

PR #74, opened 2016-02-16
Forget container id if node is unhealthy

ryanlovett:
To address:
    docker.errors.APIError: 500 Server Error: Internal Server Error ("b'Container jupyter-USER running on unhealthy node NODE'")


rgbkrk:
Excellent.

rgbkrk:
I amusingly just ran into this during a deployment against a swarm cluster.

ryanlovett:
Awesome, thanks!

PR #80, opened 2016-03-07
Skip useradd if the user exists

ryanlovett:
If the user exists and DockerSpawner.remove_containers is not set, this script will exit immediately (-e) and the user's container won't start. data-8 is running into this. Also described at https://github.com/jupyter/dockerspawner/issues/67#issuecomment-179557330.


Carreau:
LGTM.

PR #81, opened 2016-03-08
remove jovyan user from systemuser image

minrk:
it confuses things

Related to #80


RafalSkolasinski:
It would be nice if image would be usable for an user with UID=1000. Only removing ``jovyan`` doesn't seem to allow that.

PR #87, opened 2016-04-01
Extend get_env instead of _env_default in systemuserspawner.py.

kylemvz:
This will match the DockerSpawner parent class and prevent an AttributeError since _env_default is no longer defined in the parent since PR #84.

```
File "/opt/conda/lib/python3.5/site-packages/dockerspawner/systemuserspawner.py", line 102, in _env_default
   env = super(SystemUserSpawner, self)._env_default()
AttributeError: 'super' object has no attribute '_env_default'
```

### DIFF
--- a/dockerspawner/systemuserspawner.py
+++ b/dockerspawner/systemuserspawner.py
@@ -98,8 +98,8 @@ class SystemUserSpawner(DockerSpawner):
         }
         return volumes
 
-    def _env_default(self):
-        env = super(SystemUserSpawner, self)._env_default()
+    def get_env(self):
+        env = super(SystemUserSpawner, self).get_env()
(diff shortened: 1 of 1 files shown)


carolynvs:
@minrk @willingc  Are the most recent changes (such as this one and #84) to DockerSpawner ready to be released? Apologies for pestering you about releases, if I should be going about things differently just let me know!

minrk:
Just about. I was waiting for the latest docker-related PR to JupyterHub before finishing this up, but now that's landed, I should be able to do a release here today.

minrk:
dockerspawner 0.3 is up.

carolynvs:
Thank you and have a great weekend! :sunny:

PR #91, opened 2016-05-09
Adding container_port option to bind against a different port than the default (8888)

stefanoborini:
Adds the container_port traitlet and removes the hardcoded entries.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -19,6 +19,7 @@ from traitlets import (
     Dict,
     Unicode,
     Bool,
+    Int,
 )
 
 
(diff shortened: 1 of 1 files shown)


rgbkrk:
Looks good and is backwards compatible because of the default setting.

Thank you!

minrk:
Thanks! I'm curious, under what circumstances would you change the port inside the container?

stefanoborini:
@minrk To use a different container with a different port and more than one service. As far as I see, the jupyter container serves by default on 8888 and the reverse proxy is configured to bridge toward that port. I have a container who listens on 6000 and I don't want to put it on 8888.

PR #93, opened 2016-05-13
Allow SELinux volume label on volumes

robnagler:
Red Hat uses SELinux enforcing. You need to modify the SELinux context globally, or more
narrowly [add a volume label (`:Z`) to volume binds](https://docs.docker.com/engine/userguide/containers/dockervolumes/#volume-labels). docker-py does allow an arbitrary `mode` on the binds. However, you cannot specify both `ro` and `mode`.

This PR allows the following configuration for volumes:

```python
c.DockerSpawner.volumes = {
    '/nfs/{username}': {'bind': '/home/{username}', 'mode': 'Z'}
    '/other/dir': '/old/style',
}
```

In addition, this PR fixes the following issues:

1) volume_binds was formatting username on the mount_points, but volume_mount_points was not.
2) volume_mount_points should be sorted, because you might be mounting one point on another.

There's a unit test for the new code. It uses pytest, which may not be appropriate, but it's what I know. :)

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -89,10 +89,17 @@ class DockerSpawner(Spawner):
         config=True,
         help=dedent(
             """
-            Map from host file/directory to container file/directory.
-            Volumes specified here will be read/write in the container.
-            If you use {username} in the host file / directory path, it will be
-            replaced with the current user's name.
+            Map from host file/directory to container (guest) file/directory
+            mount point and (optionally) a mode. When specifying the
+            guest mount point (bind) for the volume, you may use a
+            dict or str. If a str, then the volume will default to a
+            read-write (mode="rw"). With a dict, the bind is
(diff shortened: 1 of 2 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -196,15 +203,10 @@ def volume_mount_points(self):
...
-        Returns a list of all the values in self.volumes or
+        Returns a sorted list of all the values in self.volumes or
         self.read_only_volumes.
         """
-        return list(
-            itertools.chain(
-                self.volumes.values(),
-                self.read_only_volumes.values(),
-            )
-        )
```
rgbkrk: Now that `itertools` is not used in this file, can you take out the import from above to satisfy pyflakes?

rgbkrk:
Considering we had no tests from before, I'm happy with any tests. Thank you @robnagler

PR #98, opened 2016-05-28
document overriding get_ip_and_port for network_mode=host

robnagler:
While this might be obvious to some, it took me quite a while to figure out how to do what I wanted to do. You could also use an [empty yield after the return](http://stackoverflow.com/a/13243870/3075806), but `async` is the better solution for Python 3.

I wanted to add the ability to set "network_name = host", but that's much more involved, and jupyterhub/jupyterhub#587 seems to be preferable.


minrk:
I should change this method to be a tornado coroutine, like all other async methods throughout the codebase. It's confusing for it to be unique like this (#99)

PR #102, opened 2016-06-15
Clarify language around use of DockerSpawner

yuvipanda:
There's no reason it is useful only for 'basic, temporary' servers.


minrk:
Yes, indeed. Might be a vestige of an earlier text that recommended DockerSpawner as the thing to choose for temporary/workshop deployments, but that certainly isn't the only use case.

PR #103, opened 2016-06-20
Change legacy IPython code to jupyter_client

Joshun:
juypter_client should now be used instead of ipython.utils; also adding it to the requirements


willingc:
@Joshun Thanks for submitting this pull request. I've done an initial review and the code submitted seems reasonable to me.

@minrk Would you mind taking a look at this and determine which release this might be appropriate for? Thanks.

minrk:
👍 thanks!

PR #54, opened 2015-11-17
Add support for Docker Swarm

Xarthisius:
This is a port of @jhamrick 's work in order to enable usage of Docker Swarm instead of Docker Engine.

PR #112, opened 2016-08-20
Add section to README on docker swarm

jhamrick:
So that it's in a more discoverable place rather than buried deep within #54 😄


minrk:
Yay, thanks @jhamrick !

Issue #90, opened 2016-05-02
{username} in container mount point


eschlon:
Assuming my username is *eschlon*, my goal was to mount */home/eschlon* on the host to */home/eschlon* inside the container. It seemed relatively straightforward, this seems to mostly work fine:

```
c.DockerSpawner.volumes = {'/home/{username}': '/home/{username}'}
```

However, I noticed that in addition to */home/eschlon* I ended up with an additional literal ***/home/{username}***. It seems that while `user.name` is interpolated into both the local and container paths in `volume_binds` the same is not true when local volumes are created in `volume_mount_points`.

I've overridden that method in my custom Spawner like so:
```
    @property
    def volume_mount_points(self):
        """
        Override of the DockerSpawner method.

        The parent method fails to handle cases where internal docker mount points have {username} in them. Thus
        something like {'/home/{username}': '/home/{username}'} will end up creating a literal "{username}".
        """
        return [value.format(username=self.user.name) for value in list(self.volumes.values()) +
                                                                   list(self.read_only_volumes.values())]
```

Things work as expected, though it does seem like it might properly belong here. I can submit a pull request if you think it'd be useful (or an owner can just make the change, as it's pretty small).

minrk:
@eschlon Sounds like the right fix, please do make a PR. Thanks!

PR #118, opened 2016-09-20
Fixes Issue #90 and added code to normalize username substitution

whitead:
Added and tested code as suggested in Issue #90 discussion. Found additional issue with usernames that cannot be used as filenames when inserting them into volume labels. I added slugify code from Django framework to normalize usernames so they can be filenames.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
 services:
   - docker
 before_install:
-  - pip install  -f travis-wheels/wheelhouse -v --pre -r dev-requirements.txt
+  - pip install -v --pre -r dev-requirements.txt
 install:
   - pip install -e .
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,11 @@ python:
 services:
   - docker
 before_install:
-  - pip install pyflakes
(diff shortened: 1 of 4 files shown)


rgbkrk:
`slugify` looks quite testable - would you be able to add a test?

whitead:
I mean, I guess if you want we can copy-pasta from [django](https://github.com/django/django/blob/ecb59cc6579402b68ddfd4499bf30edacf5963be/tests/template_tests/filter_tests/test_slugify.py). Do we need this?

whitead:
Hmm, I noticed the tests weren't actually being run and my code breaks the only test there is. Let me update my PR. I'll also put the test scripts into the travis file for you.

whitead:
Alright, I've written a new test, made the old ones pass, and put them into the `.travis.yml` file so they will be part of the CI.

rgbkrk:
Now that was going the extra mile. Thank you!

minrk:
Sorry, I missed this one. I think there are some issues with how slugify is being used. In particular, it's coercing valid, safe values without a mechanism to allow them, which could be problematic.

@whitead can you give an example of a username that was producing a volume-bind that doesn't work? Non-ascii values should be allowed, and case should probably be left alone, as well.

PR #119, opened 2016-09-27
don’t do an editable install on Travis

minrk:
do a real install


minrk:
Nevermind, I see it's due to coverage picking the wrong files if you install the package.

Will fix that another time.

PR #120, opened 2016-09-27
revert slugify

minrk:
it was scrubbing valid identifiers.

- non-ascii characters should be allowed in paths
- spaces should not occur in usernames, but they are valid in files, anyway
- unicode normalizing probably shouldn't occur, either

@whitead do you have any examples of the usernames that prompted slugify?

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -23,18 +23,6 @@ from traitlets import (
 )
 
 
-def slugify(value):
-    """
-    Normalizes string, converts to lowercase, removes non-alpha characters,
-    and converts spaces to hyphens.
-    """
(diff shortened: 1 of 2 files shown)


whitead:
We specifically have a deployment where usernames have domains (e.g., user@domain.com) which includes ```.``` and ```@```. Having paths containing those characters causes errors with docerkspawner, our own docker scripts and cause headaches with our backup system.

whitead:
Nope, I'm wrong there is nothing wrong with dockerspawner with the `@` and `.` It's in our start-up script.  I guess this isn't necessary. Sorry about the confusion, feel free to revert.

rgbkrk:
Thanks for checking on this and reverting Min.

PR #122, opened 2016-10-10
Add clarification on persistence as suggested by @akhmerov

willingc:
Added clarification based on Gitter discussion.

@akhmerov @minrk

### DIFF
--- a/singleuser/README.md
+++ b/singleuser/README.md
@@ -10,6 +10,23 @@ class to launch user notebook servers within docker containers.
 
 This particular server runs (within the container) as the `jupyter` user, with
 home directory at `/home/jupyter`, and the IPython example notebooks at
-`/home/jupyter/examples`. This home directory is *not* persistent by default;
-thus, this image should be used with temporary or demonstration JupyterHub
+`/home/jupyter/examples`.
+
+## Important note on persistence
+
+This home directory, `/home/jupyter`, is *not* persistent; thus,
+this image should be used with temporary or demonstration JupyterHub
 servers.
+
+If you would like persistence of data, additional configuration is
+required. See [jupyterhub/jupyterhub-deploy-docker][jupyterhub-deploy-docker]'s
+documentation on [Creating a JupyterHub data volume][data volume] and
+[How can I backup a user's notebook directory][backup user directory]
+for configuration information about persistence and mounting a docker volume.
+You may also find the [official Docker documentation on data volumes][docker data volumes] helpful.
+
+
+  [jupyterhub-deploy-docker]: (https://github.com/jupyterhub/jupyterhub-deploy-docker)
+  [data volume]: https://github.com/jupyterhub/jupyterhub-deploy-docker#create-a-jupyterhub-data-volume
+  [backup user directory]: https://github.com/jupyterhub/jupyterhub-deploy-docker#how-can-i-backup-a-users-notebook-directory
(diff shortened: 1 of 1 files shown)


akhmerov:
Reading the text again I'm actually wondering if it's the correct place for the clarification as opposed to the main [readme](https://github.com/jupyterhub/dockerspawner#dockerspawner-1). After all the image contains nothing about the volume nor about `jupyterhub_config`, so the comments seem a bit out of place here.

willingc:
@akhmerov Reading both readme files. I'm leaning a bit toward rewriting the main readme as the project has evolved over time. I think we can improve the current distinction between dockerspawner and systemuserspawner. Would you mind leaving a few bullets here on why you prefer dockerspawner over systemuserspawner?

cc/ @minrk

akhmerov:
@willingc I'm not a devops, so my opinion is definitely uneducated. Nonetheless, `DockerSpawner` requires less configuration on the host system, and therefore I believe it is easier to migrate/update/cleanly remove. Since you don't create the users on the host system, there is less concern of the host system security, and resource limits can be managed exclusively via docker tools.

One purely cosmetic advantage of `SystemUserSpawner` is due to the docker quirk that maps all internal container users on the host system users. Therefore tools like `top` would show a bunch of processes launched by a single user on the host system. Of course, `SystemUserSpawner` doesn't incur any drawbacks if it is being set up on a server that already has users created (but in that case `SudoSpawner` may be an easier choice).

singleuser/README.md
```diff
@@ -10,6 +10,23 @@ class to launch user notebook servers within docker containers.
...
 This particular server runs (within the container) as the `jupyter` user, with
 home directory at `/home/jupyter`, and the IPython example notebooks at
-`/home/jupyter/examples`. This home directory is *not* persistent by default;
-thus, this image should be used with temporary or demonstration JupyterHub
+`/home/jupyter/examples`.
+
+## Important note on persistence
+
+This home directory, `/home/jupyter`, is *not* persistent; thus,
+this image should be used with temporary or demonstration JupyterHub
```
minrk: Maybe something along the lines of "not persistent by default, so some configuration is required unless it is to be used with temporary or demonstration JupyterHub deployments"

minrk:
I think @akhmerov's point about moving more of the details about *how* to do the configuration to the main README is a good idea, and reduce the note on the image README to 'additional configuration is required' if persistence is desirable (rather than the current implication that persistence isn't achievable).

Re: heuristics for SystemUser vs Docker, the gist is: most cases should use Docker. You need a reason to choose SystemUser, and the primary ones are:

1. You are using docker just for environment management, but are running on a system where the users already have accounts and files they should be able to access from within the container.
2. You are using an external service that relies on distinct unix user ownership and permissions (nbgrader)

If neither of those applies, DockerSpawner is probably the right choice.

The simplest version of persistence to the host filesystem is to isolate users in the filesystem, but leave everything owned by the same 'actual' user with DockerSpawner. That is, using docker mounts to isolate user files, not ownership or permissions on the host.

willingc:
@minrk Updated PR with your suggested text. 

I'll work on #123 to incorporate the explanation above.

akhmerov:
Thanks, looks good to me.

willingc:
Thanks for the review @akhmerov on this pr and the other :sunny:

PR #126, opened 2016-11-08
Add ability to set memory limit for containers explicitly

yuvipanda:
Also set env variables to support memory discoverability with
https://github.com/jupyterhub/jupyterhub/issues/854


yuvipanda:
@minrk this is ready for review :) Note that I'm only adding mem_limit since I couldn't find support for cpu_limit (or cpu_shares) in http://docker-py.readthedocs.io/en/latest/api/?highlight=mem_limit create container call.

PR #131, opened 2016-12-19
indicate that we will resume if not using remove_containers

minrk:
prevents JupyterHub 0.7.1 from deleting our API tokens while we are still using them

counterpart to https://github.com/jupyterhub/jupyterhub/pull/910

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -397,6 +397,13 @@ class DockerSpawner(Spawner):
             self.log.info(
                 "Found existing container '%s' (id: %s)",
                 self.container_name, self.container_id[:7])
+            # Handle re-using API token.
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -397,6 +397,12 @@ def start(self, image=None, extra_create_kwargs=None,
             self.log.info(
                 "Found existing container '%s' (id: %s)",
                 self.container_name, self.container_id[:7])
+            # handle re-using API token. Delete the one the Hub just gave us,
+            # and restore the one we are already using if it's been deleted:
+            for line in container['Config']['Env']:
```
willingc: Thanks @minrk. I wasn't sure which container applied here.

PR #134, opened 2017-01-02
changelog for 0.6

minrk:
cc @willingc

I released JupyterHub 0.7.1 with the `will_resume` fix, and I'd like to ship this one today or tomorrow, as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,14 @@ command line for details.
 
 ## [Unreleased]
 
+## [0.6] - 2016-01-02
+
+- Add `DockerSpawner.format_volume_name` for custom naming strategies for mounted volumes.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,14 @@ command line for details.
 
 ## [Unreleased]
 
+## [0.6] - 2016-01-02
+
+- Add `DockerSpawner.format_volume_name` for custom strategies for mounted volume names.
```
willingc: @minrk I wasn't sure as this reads which 'custom strategies' that this refers to as written. Perhaps:

`for custom naming strategies for mounted volumes.`

minrk: Thanks!

PR #137, opened 2017-01-16
Modified arguments to start in SystemUserSpawner

DeepHorizons:
The start method in SystemUserSpawner did not accept the same arguments as
the start mehtod in DockerSpawner. This made subclassing SystemUserSpawner
difficult as passing arguments was not possible.

This commit changes the arguments of start to match those of the
DockerSpawner method. It also modified the call to the super, passing in the
added arguments.

This commit does change the functionality a bit, as `working_dir` is no
longer forced to be `start.homedir` and can be overwritten.

### DIFF
--- a/dockerspawner/systemuserspawner.py
+++ b/dockerspawner/systemuserspawner.py
@@ -127,9 +127,18 @@ class SystemUserSpawner(DockerSpawner):
             state['user_id'] = self.user_id
         return state
 
-    def start(self, image=None):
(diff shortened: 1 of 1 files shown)


minrk:
The start methods probably shouldn't take any arguments, since JupyterHub will never call start with any arguments.

DeepHorizons:
I believe it needs it for the same reason `DockerSpawner` needs it, sometimes it wont be directly called by JupyterHub but by an inheriting class.

PR #149, opened 2017-03-13
docker-py is now called docker

minrk:


minrk:
I noticed a few more changes. I'll cut 0.7 first, then proceed with this.

minrk:
While looking at the update, I significantly simplified some config:

- remove several separate tls config options in favor of a single `tls_config` passthrough
- enable kwargs_from_env by default and remove the config option, rather than hiding docker's own default behavior behind a switch.

PR #151, opened 2017-03-26
clean up docker images jupyterhub/systemuser and jupyterhub/singleuser

natbusa:


barrachri:
[MAINTAINER](https://docs.docker.com/engine/reference/builder/#maintainer-deprecated) is also deprecated.

PR #166, opened 2017-07-10
add string replacement for kwargs; add network alias

saulobejo:
As with the container name, add the possibility or string substitution for any kwargs for the variables username and escapedusername.

Also, add network alias using "networkalias" in extra_create_kwargs. This way you can implement:

c.DockerSpawner.extra_create_kwargs.update({ 
 "hostname": "{escapedusername}_jupyter", 
"domainname": "example.com", 
"networkalias":  "{escapedusername}_jupyter.example.com"})

creating a fully predictable and connect DNS system for your jupyter network

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -373,6 +373,20 @@ class DockerSpawner(Spawner):
                 raise
         return container
 
+    def upgrade_kwags(self, data):
+        import sys
+        if data is not None:
+            for k, v in data.items():
+                w = v.format(username=self.user.name, escapedusername=self.escaped_name)
+                sys.stderr.write("upgrade_kwags:: {}: {} -> {}\n\n".format(k, v, w))
+                data[k] = w
+        else:
+            sys.stderr.write("DATA IS NONE\n\n")
+
+        return data
+
+
+
     @gen.coroutine
     def start(self, image=None, extra_create_kwargs=None,
         extra_start_kwargs=None, extra_host_config=None):
@@ -398,6 +412,27 @@ class DockerSpawner(Spawner):
                 volumes=self.volume_mount_points,
                 name=self.container_name,
             )
+
+            self.upgrade_kwags(self.extra_create_kwargs)
+
+            #https://github.com/docker/docker-py/issues/982
+            if "networkalias" in self.extra_create_kwargs:
+                networkalias = self.extra_create_kwargs["networkalias"]
+                if (not isinstance(networkalias, list)) and isinstance(networkalias, str):
+                    networkalias = [networkalias]
+
+                del self.extra_create_kwargs["networkalias"]
+                networking_config=self.client.create_networking_config(
+                  endpoints_config={
+                    self.network_name: self.client.create_endpoint_config(
+                        aliases=networkalias,
+                      )
+                   }
+                )
+                self.extra_create_kwargs["networking_config"]=networking_config
+
+
+
             create_kwargs.update(self.extra_create_kwargs)
             if extra_create_kwargs:
                 create_kwargs.update(extra_create_kwargs)
@@ -413,6 +448,7 @@ class DockerSpawner(Spawner):
 
             if not self.use_internal_ip:
                 host_config['port_bindings'] = {self.container_port: (self.container_ip,)}
+
             host_config.update(self.extra_host_config)
             host_config.setdefault('network_mode', self.network_name)
 


dockerspawner/dockerspawner.py
```diff
@@ -398,6 +412,27 @@ def start(self, image=None, extra_create_kwargs=None,
                 volumes=self.volume_mount_points,
                 name=self.container_name,
             )
+
+            self.upgrade_kwags(self.extra_create_kwargs)
+
+            #https://github.com/docker/docker-py/issues/982
+            if "networkalias" in self.extra_create_kwargs:
```
minrk: I don't think we should support custom things like this. `extra_create_kwargs` should be passed to `create` exactly as it is. The issue linked in the above comment explains how to write the correct networking_config according to the docker API docs.

dockerspawner/dockerspawner.py
```diff
@@ -373,6 +373,20 @@ def get_container(self):
                 raise
         return container
 
+    def upgrade_kwags(self, data):
+        import sys
+        if data is not None:
+            for k, v in data.items():
+                w = v.format(username=self.user.name, escapedusername=self.escaped_name)
+                sys.stderr.write("upgrade_kwags:: {}: {} -> {}\n\n".format(k, v, w))
```
minrk: Remove this debug statement

dockerspawner/dockerspawner.py
```diff
@@ -373,6 +373,20 @@ def get_container(self):
...
+                data[k] = w
+        else:
+            sys.stderr.write("DATA IS NONE\n\n")
```
minrk: debug here, too

dockerspawner/dockerspawner.py
```diff
@@ -373,6 +373,20 @@ def get_container(self):
                 raise
         return container
 
+    def upgrade_kwags(self, data):
+        import sys
+        if data is not None:
+            for k, v in data.items():
+                w = v.format(username=self.user.name, escapedusername=self.escaped_name)
```
minrk: We use some template formatting in a few places. It would probably be better to make the existing volume_formatting use a standard template  and namespace to consolidate these things, so that every templated string gets access to the same information.

dockerspawner/dockerspawner.py
```diff
@@ -373,6 +373,20 @@ def get_container(self):
                 raise
         return container
 
+    def upgrade_kwags(self, data):
```
minrk: This would perhaps be more descriptive as `apply_templates`

saulobejo: seems like a good idea

dockerspawner/dockerspawner.py
saulobejo: I agree. 
I just couldn't spend a lot of time on it so i copied your code into mine.
A single point for applying it would be more desirable indeed

dockerspawner/dockerspawner.py
saulobejo: I understand. But there's already an exception in the code for the mem_limit and port_bindings.
Also, i needed to add the apply_templates and this is only available inside your code.

The whole goal of this change is to create subdomains to each user. this is the core of the change

saulobejo:
Fixed most comments. responded to your questions.
Thanks!

willingc:
Hi @saulobejo, Please resolve merge conflict. Thanks :sunny:

saulobejo:
Hi @willingc . corrected semantically but not tested.
thanks

minrk:
Sorry for the slow turnaround. Responding to the last comment:

> The whole goal of this change is to create subdomains to each user. this is the core of the change

If that is a goal, then it should not be done via `extra_create_kwargs`. `extra_create_kwargs` should exclusively be keyword arguments passed to `create`. If it is additional functionality not supported by `create` keyword arguments, then it should be a new, top-level configurable and handled separately, e.g.

```python
network_aliases = List(Unicode(), config=True, help="...")

...
if self.network_aliases:
    # apply template
    network_aliases = [ t.format(username=...) for t in self.network_aliases ] 
    # create network config
    networking_config = self.client.create_networking_config(
        endpoints_config={
            self.network_name: self.client.create_endpoint_config(
                aliases=self.network_aliases,
            ),
        }
    )
    # pass network config to create
    create_kwargs["networking_config"] = networking_config
```

saulobejo:
This would be a great . much more elegant

minrk:
Apologies for the incredibly long wait! I'm going through to try to catch up with stale activity on dockerspawner.

I'm going to close this in favor of #368 which is a step toward the more generic proposal of templating create/host arguments

PR #173, opened 2017-07-25
Allow override of docker client api version

christophelec:
In some cases, it is needed to override the docker API client version. For example, we can get error messages with an outdated Swarm cluster, like this :

    docker.errors.APIError: 400 Client Error: Bad Request ("b'your client is too new (API version 1.27). The newest supported API version is 1.26.0'")

This modification keeps the default behavior, but allows to override the version via client_kwargs if necessary


minrk:
Thanks!

PR #179, opened 2017-08-10
tag images with JupyterHub version on DockerHub

minrk:
generates jupyterhub/singleuser:0.7 and jupyterhub/singleuser:0.8

and picks these by default, so that the default behavior of DockerSpawner is to launch with an image that works.

Following tags are created: 0.7 (aliases: 0.7.2, latest), master (aliases: 0.8)


minrk:
Results here: https://cloud.docker.com/app/minrk/repository/docker/minrk/jupyterhub-singleuser/builds while waiting for merge

willingc:
Nice work @minrk.

PR #181, opened 2017-08-11
respect upstream env, args, ip, port

minrk:
Instead of relying on custom environment variables that images have to handle, pass existing env/argv inherited from Spawner. This allows more images to more easily be used with DockerSpawner, since they don't need to specify any handling of special environment variables set only by DockerSpawner. Instead, they only need to have jupyterhub installed and launch `jupyterhub-singleuser` as their CMD. If the images don't launch singleuser,
`c.Spawner.cmd` config is respected and can be used.

- Spawner.cmd config is now respected
- Spawner.get_args() behavior is now respected,
  and the default is to extend the image's own CMD.
  Spawner.cmd overrides the image CMD.

Several duplicate/inaccurately named arguments are deprecated, as well:

- .container_port is replaced with the existing .port, which does the same thing
  and has the extra feature of actually being able to set the port.
- .container_ip is replaced with the more accurate .host_ip
  (it didn't set the container ip, but instead the host ip
   forwarded to the container. The existing Spawner.ip sets the bind address)
- default bind address is 0.0.0.0 (preserves assumed behavior of docker-stacks without the assumption)
- deprecate .container_image in favor of less redundant .image

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -2,10 +2,11 @@
 A Spawner for JupyterHub that runs each user's server in a separate docker container
 """
 
-import string
-from textwrap import dedent
 from concurrent.futures import ThreadPoolExecutor
 from pprint import pformat
+import string
+from textwrap import dedent
+import warnings
 
 import docker
 from docker.errors import APIError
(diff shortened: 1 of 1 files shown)


willingc:
Sorry @minrk for missing this one. This should simplify things 👍

PR #195, opened 2017-08-28
Add container_name_template config

nicolas-gibelin:
Hi,

this is a little modification to allow a per user, per docker-image container name (needed when using profilespawner):

Because i want user to have multiple docker container instances with differents docker images, i added the "user_container_name" option (by default "{username}-{imagename}"), and remove "container_prefix" which is no longer needed (because "user_container_name" can be "prefix-{username}-{imagename}-suffix")

For instance, i use profilespawner and user can select "jupyterhub/systemuser", or "jupyter/r-notebook" docker image:

With this patch the user "bob" have 2 docker containers:

bob-jupyterhub_systemuser and bob-jupyter_r-notebook. No need to remove a container to use other image.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -147,19 +147,19 @@ class DockerSpawner(Spawner):
         config=True,
         help=dedent(
             """
-            Prefix for container names. See user_container_name for full container name for a particular
+            Prefix for container names. See container_name_template for full container name for a particular
             user.
             """
         )
     )
 
-    user_container_name = Unicode(
+    container_name_template = Unicode(
         "{prefix}-{username}",
         config=True,
         help=dedent(
             """
-            Name of the container: with {username}, {imagename}, {container_prefix} replacements.
-            The default user_container_name is <prefix>-<username> for backward compatibility
+            Name of the container: with {username}, {imagename}, {prefix} replacements.
+            The default container_name_template is <prefix>-<username> for backward compatibility
             """
         )
     )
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -142,13 +142,12 @@ def _container_image_changed(self, change):
         """
     )
 
-    container_prefix = Unicode(
```
minrk: Can you leave this config for backward compatibility, please?

dockerspawner/dockerspawner.py
```diff
@@ -142,13 +142,12 @@ def _container_image_changed(self, change):
...
-        "jupyter",
+    user_container_name = Unicode(
+        "{username}-{imagename}",
```
minrk: The default should probably still be `{container_prefix}-{username}`

dockerspawner/dockerspawner.py
```diff
@@ -363,7 +362,9 @@ def escaped_name(self):
 
     @property
     def container_name(self):
-        return "{}-{}".format(self.container_prefix, self.escaped_name)
+        escaped_container_image = self.image.replace("/", "_")
+        d = {'username' : self.escaped_name, 'imagename' : escaped_container_image}
```
minrk: include container_prefix here, and probably include, `server_name = getattr(self, 'name', '')` in the namespace for the new named servers feature in 0.8.

nicolas-gibelin:
Hi, regarding your comments, i change the code and test it to support the features you asked for.

That was a good idea.

Thanks.
Nicolas Gibelin

minrk:
Thanks!

Issue #215, opened 2018-01-19
Support new Docker Swarm Mode (Docker 17+)


scnerd:
It's official, [SwarmSpawner is deprecated.](https://github.com/cassinyio/SwarmSpawner/issues/29) After several days of copying every example configuration I can find, I can't get the latest SwarmSpawner to work on Docker 17+ (specifically 17.12.0-ce).

Any chance spawning Docker Services into a new-style Docker Swarm will get supported any time soon? This repository seems like the perfect place to implement such a feature (seeing as old-style swarms were already supported), now that there's no other alternative.

scnerd:
I'm currently working on implementing this in [scnerd:swarmspawner](https://github.com/scnerd/dockerspawner/tree/swarmspawner) and it looks like there's some older, possibly stale work in [yacchin1205:with-docker-services-api](https://github.com/yacchin1205/dockerspawner/tree/with-docker-services-api). Neither approach looks nearly mature, but I'm hoping to get my branch robust enough to propose a PR.

co60ca:
FYI I patched version 3906f4bebc92b383c73fb8d06c58a7c57003939a

with
```
diff --git a/dockerspawner/swarmspawner.py b/dockerspawner/swarmspawner.py
index 501047a..f853f0c 100644
--- a/dockerspawner/swarmspawner.py
+++ b/dockerspawner/swarmspawner.py
@@ -105,7 +105,7 @@ class SwarmSpawner(DockerSpawner):
                     source=host_loc,
                     type="bind",
                     read_only=vol["mode"] == "ro",
-                    driver_config=driver,
+                #    driver_config=driver,
                 )
                 for host_loc, vol in self.volume_binds.items()
             ]
@@ -126,7 +126,7 @@ class SwarmSpawner(DockerSpawner):
             "Service %s status: %s", self.service_id[:7], pformat(service_state)
         )

-        if service_state["State"] in {"running", "pending"}:
+        if service_state["State"] in {"running", "pending", "starting"}:
             return None

         else:
```
and it works for me. You'll note that this isn't a proper fix but might help someone anyway.

minrk:
Hi! I’m going through and cleaning up old/stale issues on this repo.

SwarmSpawner has received several updates related to handling service states (e.g. #296), so I think this is probably fixed. Feel free to open a new Issue if you are still having this trouble.

PR #216, opened 2018-01-30
Swarmspawner to support new-style Docker Swarm mode

scnerd:
Added a new class called Swarmspawner by duplicating Dockerspawner and modifying it where needed. Removed deprecated fields instead of adapting them. [My working sample hub config looks like the following](https://github.com/scnerd/pythonathon_docker/blob/master/jupyterhub/jupyterhub_config.py):

```python
c.JupyterHub.ip = '0.0.0.0'

c.JupyterHub.spawner_class = 'dockerspawner.SwarmSpawner'
c.SwarmSpawner.image = os.environ.get('DOCKER_NOTEBOOK_IMAGE')
c.SwarmSpawner.use_docker_client_env = True

network_name = os.environ['DOCKER_NETWORK_NAME']
c.SwarmSpawner.use_internal_ip = True
c.SwarmSpawner.network_name = network_name
c.SwarmSpawner.extra_host_config = {'network_mode': network_name}
c.SwarmSpawner.remove_services = True

notebook_dir = os.environ.get('DOCKER_NOTEBOOK_DIR') or '/home/jovyan/work'
c.SwarmSpawner.notebook_dir = notebook_dir
# c.SwarmSpawner.volumes = {'notebooks/jupyterhub-user-{username}': notebook_dir}
# c.SwarmSpawner.extra_create_kwargs.update({'volume_driver': 'local'})

c.SwarmSpawner.debug = True

c.Spawner.default_url = '/lab'
c.Spawner.args = ['--NotebookApp.allow_origin=*']

c.JupyterHub.hub_ip = '0.0.0.0'
# c.JupyterHub.hub_port = 8080

c.Spawner.cpu_limit = 0.25

c.Spawner.mem_limit = '256M'
```

Mounting volumes not yet tested. Not sure about other other features not explicitly tested in the above configuration.

To reduce the technical debt of this addition, it would probably be worthwhile extracting a superclass for both Dockerspawner and Swarmspawner (especially their shared attributes). Most of the major differences are isolated to a few functions (start, stop, poll, get_service/container, maybe one or two others).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,6 @@ env
 oauthenticator.py
 jupyterhub.sqlite
 src
+
+# Pycharm
+.idea
--- a/README.md
+++ b/README.md
@@ -30,12 +30,18 @@ pip install dockerspawner
 
 ### Choose a spawner
 
-Two basic types of spawners are available for dockerspawner:
+Three basic types of spawners are available for dockerspawner:
 
 - [DockerSpawner](#DockerSpawner): useful if you would like to spawn 
   single user notebook servers on the fly. It will take an
   authenticated user and spawn a notebook server in a Docker container
   for the user.
+
+- [SwarmSpawner](#SwarmSpawner): same behavior as DockerSpawner, but
+  launches single user notebook servers as Docker Swarm mode services
+  instead of as individual containers. This allows for running JupyerHub
+  in a swarm so that notebook containers can be run on any of multiple
+  servers.
 
 - [SystemUserSpawner](#SystemUserSpawner): useful if you would like to
   spawn single user notebook servers that correspond to the system's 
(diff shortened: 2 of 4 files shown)


scnerd:
Resolves #215

sangramga:
Wow a really needed PR. @minrk Hope it is merged soon.

sangramga:
@scnerd Just a doubt? Will ``get_ip_and_port()`` work with ``network = host`` (As mentioned in the docstring )
```
def get_ip_and_port(self):
        """Queries Docker daemon for service's IP and port.  
        If you are using network_mode=host, you will need to override  
        this method as follows::  
            @gen.coroutine  
            def get_ip_and_port(self):  
                return self.host_ip, self.port  
        You will need to make sure host_ip and port  
        are correct, which depends on the route to the service  
        and the port it `opens.`  
        """
```

minrk:
Thanks, this is great! I think we should make SwarmSpawner inherit from DockerSpawner so as to avoid reimplementing too many things, but that can be done in a separate PR.

scnerd:
I think inheritance would be great, though I think there are enough little differences in the configuration and behavior of each that I'd suggest creating a shared, abstract base class, and making both DockerSpawner and SwarmSpawner inherit from it. I meant to get around to this, but wanted to get the original SwarmSpawner tested and merged first before creating a branch that modified the main DockerSpawner.

PR #219, opened 2018-02-10
add image_whitelist and default options form

### LINKED ISSUE #236: Multi Image Docker Spawner
This is less a bug, and also not quite a pull request, but more an idea or a recipe: You can allow users to chose the docker image they are using upon starting their server by using something like the following

```python
from dockerspawner import SystemUserSpawner

class MultiDockerImageSpawner(SystemUserSpawner):
    images = {
        'SciPy': 'jupyter/scipy-notebook:0f73f7488fa0',
        'Tensorflow': 'jupyter/tensorflow-notebook:59904dd7776a',
        'R': 'jupyter/r-notebook:59904dd7776a',
    }
    def _options_form_default(self):
        outval = """
        <label for="image">Docker Image</label>
        <select name="image">
        """
        for name, image in self.images.items():
            outval += "<option value=\"%s\">%s (%s)</option>" % (name, name, image)

        outval += """
        </select>
        """
        return outval

    def options_from_form(self, formdata):
        options = {}
        options['image'] = formdata.get('image', ['SciPy'])[0]
        self.image = self.images[options['image']]
        return options
```

@minrk what do you think about this? would you be interested in wrapping this up in a class that goes in the codebase?

minrk:
if c.DockerSpawner.image_whitelist is specified, a default options form is constructed with a select dropdown for choosing the image to launch.

image_whitelist is a dict of `{key: image}`, where the key is the value to be used in the form (and shown to humans), while the values are the actual docker images. In this case, the actual image name is not shown to users.

`image_whitelist` may be specified as a list, in which case it is cast to a dict of the form `{image:image}`, where the keys are the same as the values (users see actual image names, rather than abbreviations/descriptions).

closes #236


dhirschfeld:
One thing I'd like is to be able to specify different images for different groups.

I'm not sure if there is currently any concept of a group membership for authenticated users? It's something I've been meaning to look into...

dhirschfeld:
In the config I would hard-code  the allowed images *for each group* - i.e. the `image_whitelist` would be a mapping of the group name to image names
```
c.DockerSpawner.image_whitelist = {
    prod_group : {'display_name': 'image_name'},
    uat_group : {'display_name': 'image_name'},
    dev_group : {
        'display_name1': 'image_name1',
        'display_name2': 'image_name2',
    },
}
```
The allowed images would then be chosen from the union of all groups the user was a member of and the options form would group the image names by the group name.

I understand if this is too specialised a usecase to be incorporated in DockerSpawner itself - I just wanted to put it out there as something to consider...

dhirschfeld:
> *In the config I would hard-code the allowed images for each group*

I'd ideally like this to be configurable in the JupyterHub admin UI - it would be a bit awkward to have to restart JupyterHub just to add a new image to the whitelist

minrk:
Thanks for your use case. I think I probably won't have support for group-specific image lists in the base class since I think that can take too many forms, but I'll make sure that the necessary overrides are convenient via config, so all you'll have to provide is a callable that returns the right whitelist for the current user.

As for groups, yes JupyterHub does have the concept of groups, though a user can be a member of multiple groups (like unix), so you have to be a little careful. There isn't UI for managing groups, but you can do it via REST and/or config.

Admin UI can only be used to affect values stored in the db, not config. Anything that lives only in config would get lost on Hub relaunch.

dhirschfeld:
Just for completeness' sake I'll point out that my proposal is similar/related to the below issues which are spread about different repos:
 * https://github.com/jupyterhub/wrapspawner/pull/15
 * https://github.com/jupyterhub/wrapspawner/issues/7
 * https://github.com/jupyterhub/jupyterhub/issues/1681

nils-werner:
Would it make sense to design `SystemUserSpawner`, `MultiDockerImageSpawner` etc. to be mixins, so their behaviour can be mixed and matched depending on the users needs?

minrk:
@nils-werner that's already how it works, since this functionality is defined on DockerSpawner and SwarmSpawner is a subclass, so it inherits everything from DockerSpawner.

@dhirschfeld this now allows `image_whitelist` to be a callable, so you can specify your image whitelist as a function of the user it's generating the whitelist for.

nils-werner:
> that's already how it works, since this functionality is defined on DockerSpawner and SwarmSpawner is a subclass, so it inherits everything from DockerSpawner.

I mean as mixins, not just deriving from one class. Using mixins, I could create a spawner that offers  both systemuser and swarm using

    class MySpawner(SystemUserSpawner, SwarmSpawner):
         pass

Data-drone:
@minrk is this finished now? when will it be merged into the dockerSpawner ?

oshiroscope:
Thank you for wonderful PR  @minrk. This is what I have been searching for. 
Let me ask a quetion.  How can we change the target image ? To my understanding, we have to shutdown the jupyterhub to change the image to spawn with a single account, is it right ?

minrk:
Yes, I believe this is ready.

pinsleepe:
@minrk working like a charm!

pgonyan:
Is this functionality working in 0.10?

I wrote in jupyterhub_config.py
c.DockerSpawner.image_whitelist = {'esit/r-notebook':'esit/r-notebook', 'esit/r-notebook':'esit/r-notebook'}

But... logs adive
jupyterhub | [W 2018-11-23 14:34:00.360 JupyterHub configurable:168] Config option `image_whitelist` not recognized by `DockerSpawner`.

Using Ldapauthenticator
jupyterhub | [I 2018-11-23 14:33:08.843 JupyterHub app:1667] Using Authenticator: ldapauthenticator.ldapauthenticator.LDAPAuthenticator-1.
2.2
jupyterhub | [I 2018-11-23 14:33:08.844 JupyterHub app:1667] Using Spawner: dockerspawner.dockerspawner.DockerSpawner-0.10.0

pgonyan:
Is there a documentation page for this functionality or any examples?

psychemedia:
I'm interested for docs on this too...

eg I'm using a simple `Dummyauthenticator` but I'm not seeing the image selection list for a pre-existing user?

psychemedia:
@pinsleepe Yep, I've been using installs from the Guthub masters.

dhirschfeld:
I'm keen to use this functionality but need a release to be able to do so easily

flixr:
Dito.
Is there a timeline for the next release?

minrk:
Working on the 0.11 release now.

MaisamMD:
How can map a specific volume to each of the images in the white list? 
something like:
c.DockerSpawner.volumes = {'host/directory':'/container/directory'}

PR #230, opened 2018-04-09
Remove deprecated extra_start_kwargs

### LINKED ISSUE #192: Remove extra_start_kwargs config?
`docker-py` raises an exception if you pass any extra args or kwargs to the start method:

https://github.com/docker/docker-py/blob/89195146adde47181156b0c204c72b790c69cf85/docker/api/container.py#L1073-L1078

...which then causes the spawn of the container to fail with the message:
```
Unhandled error starting dhirschfeld's server: Providing configuration in the start() method is no longer supported. Use the host_config param in create_container instead.
```

Given that it's a hard fail perhaps it's best to remove `extra_start_kwargs` from `DockerSpawner` so that the user can't do something which has no possibility of succeeding.

dhirschfeld:
Passing kwargs to start now raises an exception in docker-py

Closes #192

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -72,7 +72,7 @@ class DockerSpawner(Spawner):
         self._user_set_cmd = True
 
     container_id = Unicode()
-    
+
     # deprecate misleading container_ip, since
     # it is not the ip in the container,
     # but the host ip of the port forwarded to the container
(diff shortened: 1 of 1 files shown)


dhirschfeld:
Since [3 years ago](https://github.com/docker/docker-py/commit/5eacb986d706226ccfd9a6b1021828b6eb67dbad) the docker library has raised an Exception if any args or kwargs are passed to start.

image

It's probably time to remove this feature which no longer works from DockerSpawner.

PR #235, opened 2018-06-05
Fixing README's subsection links

ionlights:
Each spawner's hyperlinks didn't properly jump, could be due to a change in Markdown parsing; but all's resolved now. :smiley: 

Hyperlinks fixed: 
- DockerSpawner 
- SwarmSpawner 
- SystemUserSpawner


minrk:
Thanks!

PR #265, opened 2018-10-11
Wrong documentation: the correct property name should be container_im…

xct24:
…age rather than image


minrk:
Thanks for the PR! But this change is incorrect. `container_image` is a deprecated alias for `image`. `image` is the right name.

PR #269, opened 2018-10-23
Return None on poll of "starting" container

vilhelmen:
[Hub docs](https://jupyterhub.readthedocs.io/en/stable/api/spawner.html#jupyterhub.spawner.Spawner.poll) say that poll should return None when the server is still starting.

DockerSpawner's poll function most likely needs to change as well, but I don't really understand how it's doing the status check, so I'm not going to touch it.


minrk:
Thanks! I'm not entirely sure if DockerSpawner does need to change, since containers have fewer states than services. But we should indeed double-check.

Issue #264, opened 2018-10-11
SwarmSpawner: Run single-user containers only on worker nodes


farzadz:
I was trying to make single-user servers run on nodes other than Jupyterhub's node. Is there any type of config for swarm spawner to specify constraints/filters about the placement of servers?

minrk:
closed by #270

PR #270, opened 2018-10-25
docker service placement constraints as config option

wlohman-surfsara:
I was having the same question as #264 

Added placement as a config option, so that this works:
`c.SwarmSpawner.extra_placement_spec = { 'constraints' : ['node.role==worker'] }`


minrk:
Thanks!

PR #288, opened 2019-02-22
fix doc of stop_object and describe remove better

robnagler:


willingc:
Thanks @robnagler for improving these docstrings. Much more helpful. :cookie:

Issue #299, opened 2019-04-18
SystemUserSpawner: Derive host homedir from passwd entry


AdrianoKF:
## Scenario

I'm using `SystemUserSpawner` together with `PAMAuthenticator` to provide LDAP-based logins and mount user home directories as Docker volumes.

However, I would prefer not to mix LDAP and local user home directories, and have sequestered them into `/home` for local and `/opt/data` for remote LDAP users on my server.

However, `SystemUserSpawner` does not read the host user home directories from the passwd entry for the user (as it already does for the UID), but rather relies on the template in `host_homedir_format_string`. Therefore, the scenario described above cannot be implemented without changes to the file system, e.g., by creating symlinks for the local users under `/opt/data`.

## Suggested Solution
Would it make sense to fetch this information via `pwd.getpwnam()` instead, and allow the template as a fallback solution? One possible downside is that the `pwd` package breaks platform-independence for non-Unix systems - however since it is been used in [_user_id_default()](https://github.com/jupyterhub/dockerspawner/blob/9c1efa0e396e4f2b9939fde7dacbe4187ceb4041/dockerspawner/systemuserspawner.py#L108) already, that might be a non-issue.

### Code changes
I have already implemented a quick-and-dirty fix, which I can provide as a pull request, if needed. In `dockerspawner/systemuserspawner.py` I have replaced [host_homedir()](https://github.com/jupyterhub/dockerspawner/blob/9c1efa0e396e4f2b9939fde7dacbe4187ceb4041/dockerspawner/systemuserspawner.py#L51) with the following implementation:

```python
@property
def host_homedir(self):
    """
    Path to the volume containing the user's home directory on the host.
    Looked up from pwd if an empty format string has been specified.
    """
    if self.host_homedir_format_string != "":
        homedir = self.host_homedir_format_string.format(username=self.user.name)
    else:
        import pwd
        homedir = pwd.getpwnam(self.user.name).pw_dir

    self.log.info(f"Using system user home directory: {homedir}")
    return homedir
```

minrk:
There have been cases where user_id is set via the Authenticator, such that `pwd` never gets called. My memory is hazy, but I believe this was in a distributed cluster with home directories on NFS where users wouldn't necessarily be found by pwd lookup.

Gating this behind an option should be safe, though, e.g.:

```python
use_pwd_homedir = Bool(False, config=True, help="Use pwd to lookup home directories")
```

[...]

PR #302, opened 2019-04-26
Allow lookup of host homedir via `pwd`

AdrianoKF:
Fixes issue jupyterhub/dockerspawner#299.
The home directory of the user on the host can be looked up via the
`pwd` database, which is useful when dealing with non-local users
(e.g., through LDAP authentication).

The lookup can be enabled by setting `host_homedir_format_string` to
an empty string or `None`.


minrk:
Thanks! Sorry for leaving you hanging

PR #304, opened 2019-05-08
SystemUserSpawner: Pass group id to the container

zeehio:
Thanks for your work.

The `SystemUserSpawner` allows to span a docker container with the `NB_UID` environment variable set, so files are created with the username we want (mapped from the system). However it does not pass the `NB_GID` variable, so the group ID is always fixed to 100 or to any other number.

This PR provides the `group_id` feature and passes it to `NB_GID` so the created single user notebooks know with which group ID files should be created.

### DIFF
--- a/dockerspawner/systemuserspawner.py
+++ b/dockerspawner/systemuserspawner.py
@@ -50,6 +50,20 @@ class SystemUserSpawner(DockerSpawner):
         )
     )
 
+    group_id = Integer(-1,
+        help=dedent(
+            """
+            If system users are being used, then we need to know their group id
+            in order to mount the home directory with correct group permissions.
+
+            Group IDs are looked up in two ways:
+
+            1. stored in the state dict (authenticator can write here)
+            2. lookup via pwd
+            """
+        )
+    )
+
     @property
     def host_homedir(self):
         """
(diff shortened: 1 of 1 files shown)


dockerspawner/systemuserspawner.py
```diff
@@ -101,6 +115,7 @@ def get_env(self):
             NB_USER=self.user.name,
             USER_ID=self.user_id, # deprecated
             NB_UID=self.user_id,
+            NB_GID=self.group_id,
```
minrk: This probably doesn't want to be set `if self.group_id < 0`

zeehio: I added a check for that, I hope it's what you wanted.

https://github.com/jupyterhub/dockerspawner/pull/304/commits/10183dcab732b1cc2c1c999ccbafe29cf4aeaf79

dockerspawner/systemuserspawner.py
```diff
@@ -112,6 +126,8 @@ def get_env(self):
             NB_UID=self.user_id,
             HOME=self.homedir,
         ))
+        if self.group_id >= 0:
+            env.update(dict(NB_GID=self.group_id))
```
GeorgianaElena, line 130:
```suggestion
            env.update(NB_GID=self.group_id)
```

PR #310, opened 2019-05-22
fix cpu_limit no effect bug

CandleSense:
In the process of using DockerSpawner, I want to limit the CPU usage of the entire container, but cpu_limit only limits the CPU usage of each ipynb, 
image

image

so I want to use c.dockerspawner. extra_host_config to limit that, 
```
c.DockerSpawner.extra_host_config.update({
    'cpu_period': 100000,
    'cpu_quota': 50000,
})
```
and I find that it doesn't work. So now I've made some changes

```
 $ docker inspect jupyter-kkk|grep Cpu
            "CpuShares": 0,
            "NanoCpus": 0,
            "CpuPeriod": 100000,
            "CpuQuota": 50000,
            "CpuRealtimePeriod": 0,
            "CpuRealtimeRuntime": 0,
            "CpusetCpus": "",
            "CpusetMems": "",
            "CpuCount": 0,
            "CpuPercent": 0,
```

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -858,7 +858,8 @@ class DockerSpawner(Spawner):
 
         # build the dictionary of keyword arguments for host_config
         host_config = dict(binds=self.volume_binds, links=self.links)
-
+        self.log.debug('extra_host_config: %s', self.extra_host_config)
+        host_config.update(self.extra_host_config)
         if getattr(self, "mem_limit", None) is not None:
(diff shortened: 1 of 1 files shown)


CandleSense:
Although I can solve it in the following way, I think it is not a good way to achieve it, because it bypasses `docker.APIClient.create_host_config(**host_config)`

```
c.DockerSpawner.extra_create_kwargs.update({
                                            'host_config': {
                                                'CpuPeriod': 100000,
                                                'CpuQuota': 50000,
                                            },
})
```

Issue #257, opened 2018-08-27
single user multiple containers?


Data-drone:
can a single user get multiple containers?

I have a requirement for users to have access to multiple environments with sometimes clashing library versions to run comparisons

minrk:
You can use the options_form to allow users to pick from a collection of docker images

zeehio:
This feature (allowing a single user simultaneously have servers with multiple containers) requires multiple pieces:

- The jupyterhub part has been implemented at https://github.com/jupyterhub/jupyterhub/issues/766 and it can enabled with `c.JupyterHub.allow_named_servers=True`.
- When that setting is enabled, `/hub/home` allows you to create new servers with different spawner options (different images).
- However I guess dockerspawner does not implement its part yet. Only one container is created and a 404 error is obtained when I try to access other servers.

zeehio:
It works!

```
c.JupyterHub.allow_named_servers=True
c.SystemUserSpawner.name_template = '{prefix}-{username}-{servername}'
```

The default `name_template` is bad for a multiple named servers configuration.

PR #315, opened 2019-06-12
Make default name_template compatible with named servers.

danielballan:
As noted by @zeehio in [this comment](https://github.com/jupyterhub/dockerspawner/issues/257#issuecomment-492275806)
the default ``name_template`` is not compatible with using named servers. Should the default be updated?

The current docstring mentions backward compatibility. If
backward-compatibility of container names is very important, perhaps it would
be better to leave the default as is and simply mention ``{servername}`` in the
docstring and the README.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -306,7 +306,6 @@ class DockerSpawner(Spawner):
     )
 
     name_template = Unicode(
-        "{prefix}-{username}-{servername}",
         config=True,
         help=dedent(
             """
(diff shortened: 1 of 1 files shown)


danielballan:
Another possibility would be to make the default template

```
{prefix}-{username}{servername}
```

which means that the default (unnamed) server would create a container of the same name as before, retaining back-compat. The disadvantage, of course, is that the username runs right into the server name without a separator.

minrk:
For compatibility, using a dynamic default should do it without breaking the default case:

```python
@default('name_template')
def _default_name_template(self):
    if self.name:
        return "{prefix}-{username}-{servername}"
    else:
        return "{prefix}-{username}"
```

danielballan:
Oh, perfect. At the time I submitted this PR I hadn't yet discovered dynamic defaults. Really nice solution. Pushed.

danielballan:
One of the three builds failed---see [build log](https://travis-ci.org/jupyterhub/dockerspawner/jobs/593527762?utm_medium=notification&utm_source=github_status). It looks like a transient failure to me.

danielballan:
A friendly bump on this. Do you think the CI failures are real?

manics:
I've closed/reopened to trigger an updated CI check

GeorgianaElena:
The CI failures from now are definitely unrelated.
When @manics closed/reopened the PR, [the tests passed](https://travis-ci.org/github/jupyterhub/dockerspawner/builds/593527759), but I've restarted the build and now they [fail again because of a version conflict](https://travis-ci.org/github/jupyterhub/dockerspawner/builds/698951135).

I believe this LGTM :tada: What do you think, @manics?

manics:
Thanks @danielballan , sorry for the delay in getting this merged

danielballan:
Not a problem! Thanks for your work on maintaining JupyterHub!

PR #318, opened 2019-06-21
Add auto_remove to host_config

jtpio:
Add support for the `auto_remove` option in `host_config`, to [enable auto-removal of the container on daemon side when the container’s process exits](https://docker-py.readthedocs.io/en/stable/api.html#docker.api.container.ContainerApiMixin.create_host_config):

- Cover the case of the main process (pid 1) terminating itself, for example when using `--NotebookApp.shutdown_no_activity_timeout`
- Reuse the existing `remove` option

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -969,7 +969,10 @@ class DockerSpawner(Spawner):
 
         # build the dictionary of keyword arguments for host_config
         host_config = dict(
-            binds=self.volume_binds, mounts=self.mount_binds, links=self.links
+            auto_remove=self.remove,
+            binds=self.volume_binds,
+            links=self.links,
+            mounts=self.mount_binds,
         )
 
         if getattr(self, "mem_limit", None) is not None:
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -1,6 +1,7 @@
 """
 A Spawner for JupyterHub that runs each user's server in a separate docker container
 """
+import asyncio
 import os
 import string
 import warnings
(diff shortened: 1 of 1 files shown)


jtpio:
Thanks @GeorgianaElena for the review and the suggestions :+1: 

> But the idea behind this PR is to have it tied to the remove option (because this is what one would expect to happen when remove is True), right?

Yes the idea was to have the behavior described in the `help` docstring for the `remove` option:

https://github.com/jupyterhub/dockerspawner/blob/855665fcc7182ef8abd54424f987d81f8db358be/dockerspawner/dockerspawner.py#L511-L519

The sentence `If True, delete containers when servers are stopped` sounds like the Docker containers should be removed when they are stopped (similar to the `--rm` command line option).

But then there is also the following:

https://github.com/jupyterhub/dockerspawner/blob/855665fcc7182ef8abd54424f987d81f8db358be/dockerspawner/dockerspawner.py#L988-L990

Which seems to describe a different behavior than `auto_remove` host config option.

So yes the best for now would probably be to document `host_config` and keep the `remove` option separate?

Or maybe rephrase the docstrings to make the distinction clearer?

manics:
@jtpio Thanks for digging out those comments, I think remove==auto-remove makes sense, especially as there's this comment lower down:
https://github.com/jupyterhub/dockerspawner/blob/855665fcc7182ef8abd54424f987d81f8db358be/dockerspawner/dockerspawner.py#L1019-L1026

minrk:
I think this makes sense, auto_remove is indeed the behavior we want `remove` to have. Tests are failing because the explicit remove is still here, but it's failing with 404. I'm not sure if we should remove the explicit delete call we have or not (I'd lean toward not), but we should probably catch and suppress 404 errors when trying to delete something that may have already been deleted.

minrk:
Thanks! sorry for the long wait

PR #337, opened 2019-10-08
add `_` as safe caracter since is use to escape caracter

BobCashStory:


minrk:
Apologies for the incredibly long wait! I'm going through to try to catch up with stale activity on dockerspawner.

This won't achieve the desired result, because the escape character is explicitly removed from safe characters. From the escapism [docs](https://github.com/minrk/escapism#usage):

> escape_char will never be considered a safe value.

It can be allowed *if* you also opt-in to `allow_collisions`. However, allowing the escape character to not be escaped again would mean that maliciously crafted usernames can actually use the same container name as other users. See for example:

```python
In [15]: escapism.escape("user-name", safe=string.ascii_letters + string.digits)
Out[15]: 'user_2Dname'

In [16]: escapism.escape("user_2Dname", safe=string.ascii_letters + string.digits + "_")
<ipython-input-16-43a63e0a2a71>:1: RuntimeWarning: Escape character '_' cannot be a safe character. Set allow_collisions=True if you want to allow ambiguous escaped strings.
  escapism.escape("user_2Dname", safe=string.ascii_letters + string.digits + "_")
Out[16]: 'user_5F2Dname'

In [17]: escapism.escape("user_2Dname", safe=string.ascii_letters + string.digits + "_", allow_collisions=True)
Out[17]: 'user_2Dname' # uh oh! That's a different username with the same container name...
```

I'm going to close this, but feel free to open an Issue or another PR to discuss more specific requirements for why this is requested.

PR #338, opened 2019-10-09
Add support for specifying volume drivers

diurnalist:
**This is a draft**. But, the code is more or less in-place. See TODOs at the bottom.

This adds support for specifying a volume driver as part of the `volumes` configuration. This can be useful if you are using a volume driver that manages the volumes remotely or want to simply pass in driver options. The `local` driver (default) for example can take most args that `mount` can take, and can thus be used to mount volumes via NFS.

Example configuration:
```python
c.DockerSpawner.volumes = {
  # Old syntax still supported
  '{prefix}-{username}': '/home/{username}',
  # Example specifying NFS options
  # (https://docs.docker.com/engine/reference/commandline/volume_create/)
  'user-share': {
    'target': '/share',
    'driver': 'local',
    'driver_opts': {
      'type': 'nfs',
      'o': 'addr=192.168.1.1,rw',
      'device': ':/path/to/dir',
    },
  },
}
```

---

### TODOs

- [ ] Enable using tokens (`{username}`) as part of driver options?
- [ ] Ensure tests run
- [ ] Add test for this functionality
- [ ] Update docs

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -3,6 +3,7 @@ A Spawner for JupyterHub that runs each user's server in a separate docker conta
 """
 
 from concurrent.futures import ThreadPoolExecutor
+from copy import copy
 from io import BytesIO
 import os
 from pprint import pformat
@@ -635,7 +636,7 @@ class DockerSpawner(Spawner):
         Returns a sorted list of all the values in self.volumes or
         self.read_only_volumes.
         """
-        return sorted([value["bind"] for value in self.volume_binds.values()])
+        return sorted([v["target"] for v in self._format_volumes(self.volumes)])
 
     @property
     def volume_binds(self):
@@ -650,7 +651,8 @@ class DockerSpawner(Spawner):
         mode may be 'ro', 'rw', 'z', or 'Z'.
 
         """
-        binds = self._volumes_to_binds(self.volumes, {})
+        binds = self._volumes_to_binds(self.volumes)
+
         read_only_volumes = {}
         # FIXME: replace getattr with self.internal_ssl
         # when minimum jupyterhub is 1.0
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -1118,22 +1130,73 @@ def stop(self, now=False):
...
+        formatted = []
         for k, v in volumes.items():
-            m = mode
             if isinstance(v, dict):
-                if "mode" in v:
-                    m = v["mode"]
-                v = v["bind"]
-            binds[_fmt(k)] = {"bind": _fmt(v), "mode": m}
-        return binds
+                v = copy(v)
```
diurnalist: Note to self: this will have to be a `deepcopy` if we allow formatting in the driver options.

dockerspawner/dockerspawner.py
```diff
@@ -1118,22 +1130,73 @@ def stop(self, now=False):
...
+            else:
+                # support simple format like {source: target}
+                v = {'target': v}
+
+            if not 'target' in v:
```
diurnalist: Err, looks like I forgot to keep supporting the format `{"bind": .., ..}`--will update.

minrk:
Sorry for the incredibly long wait! Thanks for the PR, I am going through and trying to cleanup some stale activity.

As of #406, the full [mounts spec](https://docker-py.readthedocs.io/en/stable/api.html?highlight=mount#docker.types.Mount) can be used, so this feature should be available now.

diurnalist:
@minrk thanks, I'll look at adapting my code to match that one :)

minrk:
Great! Pease feel free to open a new issue/PR if I'm wrong that the mounts option covers this case. Trying to do better at staying on top of things.

Issue #262, opened 2018-09-21
SwarmSpawner unable to contact notebook container when container name has upper case characters


hpbieker:
I am using Keycloak for auth, and it gives user names that contains lower case, upper case, and numbers. By default, SwarmSpawner (and DockerSpawner) creates services (or containers) with a name in the format "<prefix>-<username>". 

The swarm load balancer is unable to do DNS lookup services that are upper case (https://github.com/moby/moby/issues/21169 / https://github.com/docker/libnetwork/issues/993) and will just hang.

It would be nice if:
* The username was automatically changed to lower case
* The container_name was automatically changed to lower case
* An error was reported if the container name ended up as an invalid name
* If docker supported DNS lookup of services in upper case...

minrk:
Which Authenticator class & config are you using? Authenticators have a `.normalize_username` method which can do things like cast-lowercase. Indeed, it is the [default behavior](https://github.com/jupyterhub/jupyterhub/blob/0.9.4/jupyterhub/auth.py#L197) to do that in most authenticators, but you have to be sure that casting to lowercase cannot result in collisions (MediaWiki being one notable example where Min and min can be different users).

DockerSpawner should indeed be ensuring that containers get valid names. It can do this with escapes, but to be safe from potential collisions, it would have to use a safe escaping mechanism, which would result in `Abc` turning into something like `-41bc` rather than `abc`. The very least we could do is raise a more informative error.

hpbieker:
I used the SwarmSpawner and OauthAuthenticator. We also use KeyCloak which generates these user names.

cblomart:
I tested with Oauthenticator (azure) and SwarmSpawner. As we have long names in the directory, i moved to using the unique_name claim which will basicaly provide the mail address of the user.

The mail addresses are ``{firstname}.{lastname}@{domain}.{ext}``.

The service names provided by SwarmSpawner are in the form ``jupyter-{firstname}_2E{lastname}_40{domain}_2E{ext}``.

The jupyter hub registers to the proxy the name in the following form ``jupyter-{firstname}_2e{lastname}_40{domain}_2e{ext}``.

Thus the resulting service cannot be resolved: the names are not the same.

Obviously someone is lowering usernames somwhere... but SwarmSpwaner is not.

Looking at my jupyter logs:
```text
[...]

PR #342, opened 2019-10-16
Add option to lowercase object name

cblomart:
In some cases services are created with upper case letter in their name but proxy knows the name in lower case. 
This adds an option to set the name in lowercase.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -623,6 +623,19 @@ class DockerSpawner(Spawner):
         Similar to using 'docker exec'
         """
     )
+    
+    lowercase_service = Bool(
+        False,
+        config=True,
+        help=dedent(
+            """
+            Lowercase service name
+            
+            Set the generated service name to lowercase to avoid conflict in certain situations.
+            This is disabled by default to avoid duplicate usernames.
+            """
+        ),
+    )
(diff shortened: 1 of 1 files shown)


cblomart:
Woild fix #262

dockerspawner/dockerspawner.py
```diff
@@ -623,6 +623,19 @@ def _default_use_hostname(self):
...
+    
+    lowercase_service = Bool(
+        False,
+        config=True,
+        help=dedent(
+            """
+            Lowercase service name
+            
+            Set the generated service name to lowercase to avoid conflict in certain situations.
+            This is disabled by default to avoid duplicate usernames.
```
muru: Duplicate usernames or service names?

cblomart: It actually lowercase the service name (object_name).

The main source of uniqueness is the username as the default template is “{prefix}-{username}”. So lowercasing can lead to duplication if username uniqueness depends on casing.

The question of username casing is also handled by some Authenticator via the function normalize_username.

This to explain why I chose to write about username and not service name.

minrk:
#414 makes escaping always DNS-safe

minrk:
Closing this one in favor of more general #414

PR #361, opened 2020-03-24
Update and not overwrite environment variables when using DockerSpawner

missingcharacter:
### Current behavior:

I add environment variables:

```
  'extra_create_kwargs': {
    'environment': {
      'SOMETHING': 'somevalue',
      'ANOTHERTHING': 'anothervalue'
    }
  }
```
 and this causes `JUPYTERHUB_API_TOKEN` and `JPY_API_TOKEN` to be overwritten.

### Expected behavior:

I add environment variables and these are only updated not overwritten.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -639,7 +639,7 @@ class DockerSpawner(Spawner):
             'cmd': self.post_start_cmd,
             'container': container_id
         }
-        
+
         exec_id = yield self.docker("exec_create", **exec_kwargs)
 
         return self.docker("exec_start", exec_id=exec_id)
@@ -888,7 +888,14 @@ class DockerSpawner(Spawner):
         # ensure internal port is exposed
         create_kwargs["ports"] = {"%i/tcp" % self.port: None}
 
-        create_kwargs.update(self.extra_create_kwargs)
+        # ensure we only update environment variables and not overwrite them
+        extra_create_kwargs = self.extra_create_kwargs
+        extra_environment = extra_create_kwargs.get('environment')
+
+        if extra_environment:
+          extra_create_kwargs['environment'] = {**create_kwargs['environment'], **extra_environment}
+
+        create_kwargs.update(extra_create_kwargs)
 
         # build the dictionary of keyword arguments for host_config
         host_config = dict(binds=self.volume_binds, links=self.links)
@@ -941,7 +948,7 @@ class DockerSpawner(Spawner):
         """
         # docker wants to split repo:tag
         # the part split("/")[-1] allows having an image from a custom repo
-        # with port but without tag. For example: my.docker.repo:51150/foo would not 
+        # with port but without tag. For example: my.docker.repo:51150/foo would not
         # pass this test, resulting in image=my.docker.repo:51150/foo and tag=latest
         if ':' in image.split("/")[-1]:
             # rsplit splits from right to left, allowing to have a custom image repo with port


missingcharacter:
Tests passed https://travis-ci.org/github/jupyterhub/dockerspawner/builds/666457221?utm_medium=notification&utm_source=github_status

manics:
The parent class has an `environment` property:
https://github.com/jupyterhub/jupyterhub/blob/53d76ad3a21541b0f49f2857f32c485407381949/jupyterhub/spawner.py#L422-L440

Did that not work for you?

missingcharacter:
@manics I'm using https://github.com/jupyterhub/wrapspawner and I need to declare environment variables with different values depending on the Spawner I'm using. that is why I'm doing it here

missingcharacter:
@manics do you have another opinion/observation about this PR?

missingcharacter:
@manics is there something else I should change?

minrk:
what about WrapSpawner makes setting the standard `environment` configuration not work? Can we get more detail on that? Does WrapSpawner not support per-spawner values?

missingcharacter:
@minrk I'm using WrapSpawner and my problems are:

```
JUPYTERHUB_ACTIVITY_URL
JUPYTERHUB_API_URL
```

I'm connected via VPN to multiple environments where I want to be able to spawn notebooks at, the jupyterhub has different IP addresses depending on the environment (due to the VPN), I understand I should be able to give the same hostname to the machine across different environments but I do not have access to that at the moment.

minrk:
I guess my question is, why is setting the per-spawner env in the profile `extra_create_kwargs` preferable to setting per-spawner env in the profile `environment` config?

For example:

```python
c.ProfileSpawner.profiles = [
    (
        "Docker Spawner",
        "docker",
        DockerSpawner,
        {"environment": {"JUPYTERHUB_API_URL": "https://custom.biz/hub/api"}},
    )
]
```

missingcharacter:
@minrk I tried it just like you suggested it:

```python
c.ProfileSpawner.profiles = [
    (
        "Docker Spawner",
        "docker",
        DockerSpawner,
        {"environment": {"JUPYTERHUB_API_URL": "http://<myhostname>/hub/api"}},
    )
]
```

and my values are overwritten, but with my changes and using `extra_create_kwargs` it works

yuvipanda:
Is this specific to JUPYTERHUB_API_URL? That's hard coded [here](https://github.com/jupyterhub/jupyterhub/blob/master/jupyterhub/spawner.py#L770) and might be causing the behavior you are seeing.

yuvipanda:
May I also ask why you are overriding JUPYTERHUB_API_URL?

minrk:
Aha, that makes sense! `Spawner.environment` config has lower priority than 'default' env like JUPYTERHUB_API_URL, making it not overrideable. This is changed in https://github.com/jupyterhub/jupyterhub/pull/3081 allowing environment from config to take top priority. I'd prefer to fix this issue in the root rather than allow multiple ways to set the same configuration which happen to be evaluated at different times. Does that sound okay, @missingcharacter?

@yuvipanda I think the use case was answered a few comments back:

> I'm connected via VPN to multiple environments where I want to be able to spawn notebooks at, the jupyterhub has different IP addresses depending on the environment (due to the VPN), I understand I should be able to give the same hostname to the machine across different environments but I do not have access to that at the moment.

so hub_connect_url is not uniform across spawners. I think this is an advanced case where some escape hatch is appropriate, but hopefully https://github.com/jupyterhub/jupyterhub/pull/3081 should do it.

yuvipanda:
Aaah, I missed that. Thanks, @minrk!

yuvipanda:
@missingcharacter I just merged https://github.com/jupyterhub/jupyterhub/pull/3081. Can you test with that?

missingcharacter:
Thanks @minrk and @yuvipanda , that did it!!!

Issue #367, opened 2020-04-10
custom container labels


devnull-mr:
Hi,

I am using an external `Traefik` proxy with enabled Docker provider for configuration discovery, so I need to set some custom labels in the spawned containers to make routing work correctly.

So I am trying to use `{username}` as that is supported for volumes and also used in the jupyterhub/jupyterhub-deploy-docker repo [here](https://github.com/jupyterhub/jupyterhub-deploy-docker/blob/16a970ec6b376071ba1b520e094366c49126b6d1/jupyterhub_config.py#L38).

My config is:
```
c.DockerSpawner.extra_create_kwargs.update({
  'labels': {
    'traefik.enable': 'true',
    'traefik.http.routers.{username}.entryPoints': 'ssl',
    'traefik.http.routers.{username}.rule': 'PathPrefix(`/user/{username}`)',
    'traefik.http.routers.{username}.tls': ''
  }
})
```
However, that does not work. The labels are added to the container, but `{username}` is not replaced by the username for which the container was spawned.

Is the substitution only supported for certain features (like volumes) and not supported in general ?
Is there another (better) to get the username added to the container labels ?

devnull-mr:
Any comments on this ?

I had submitted PR #368 with a possible fix and would appreciate if that could be reviewed and eventually merged.

PR #368, opened 2020-04-13
Add template value substitution for extra_create_kwargs

devnull-mr:
Fix for #367 

I have tried to mimic what is already done for volumes in `_volumes_to_binds()`.
Open for comments / enhancements of course !

Works for me locally. All values defined in `template_namespace()` are successfully replaced in the extra kwargs.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -888,7 +888,7 @@ class DockerSpawner(Spawner):
         # ensure internal port is exposed
         create_kwargs["ports"] = {"%i/tcp" % self.port: None}
 
-        create_kwargs.update(self.extra_create_kwargs)
+        create_kwargs.update(self._extra_create_kwargs_subs())
 
         # build the dictionary of keyword arguments for host_config
         host_config = dict(binds=self.volume_binds, links=self.links)
@@ -928,7 +928,6 @@ class DockerSpawner(Spawner):
         e.g. calling `docker stop`. Does not remove the container.
         """
         return self.docker("stop", self.container_id)
-
 
     @gen.coroutine
     def pull_image(self, image):
@@ -1173,3 +1172,22 @@ class DockerSpawner(Spawner):
                 v = v["bind"]
             binds[_fmt(k)] = {"bind": _fmt(v), "mode": m}
         return binds
+
+    def _extra_create_kwargs_subs(self):
+        """
+        Substitutes values from template_namespace() for extra docker create kwargs.
+        """
+        subs = {}
+
+        def _fmt(v):
+            return v.format(**self.template_namespace())
+
+        for k, v in self.extra_create_kwargs.items():
+            if isinstance(v, dict):
+                subs[_fmt(k)] = {}
+                for k1, v1 in v.items():
+                    subs[_fmt(k)].update({_fmt(k1): _fmt(v1)})
+            else:
+                subs[_fmt(k)] = _fmt(v)
+
+        return subs


devnull-mr:
Any feedback on this PR after all ?

jterstriep:
I am in the same situation as @devnull-mr and would like to take advantage of the changes. Can the Jupyterhub team provide some feedback?

minrk:
Sorry for this repo getting neglected, I'm trying to pick it up now. This logic looks just fine to me, but I would add one note that it explicitly supports formatting only two levels deep. That may be sufficient, but a general walk of lists and dicts may be preferable to be more general, plus applying the same logic to both host and create kwargs would make sense.

Still, this is an improvement on its own. I'll merge with conflicts resolved once CI is back alive after #407

dockerspawner/dockerspawner.py
```diff
@@ -1173,3 +1172,22 @@ def _fmt(v):
...
+        def _fmt(v):
+            return v.format(**self.template_namespace())
+
+        for k, v in self.extra_create_kwargs.items():
+            if isinstance(v, dict):
+                subs[_fmt(k)] = {}
+                for k1, v1 in v.items():
+                    subs[_fmt(k)].update({_fmt(k1): _fmt(v1)})
+            else:
+                subs[_fmt(k)] = _fmt(v)
```
minrk, line 1191: This will error if `v` is a list or anything other than a dict or string

dockerspawner/dockerspawner.py
```diff
@@ -1173,3 +1172,22 @@ def _fmt(v):
...
+        subs = {}
+
+        def _fmt(v):
+            return v.format(**self.template_namespace())
+
+        for k, v in self.extra_create_kwargs.items():
+            if isinstance(v, dict):
+                subs[_fmt(k)] = {}
+                for k1, v1 in v.items():
+                    subs[_fmt(k)].update({_fmt(k1): _fmt(v1)})
```
minrk, line 1189: this will error if `v1` is not a string. It could itself be a list or dict or even a `docker.types` model.

dockerspawner/dockerspawner.py
```diff
@@ -888,7 +888,7 @@ def create_object(self):
         # ensure internal port is exposed
         create_kwargs["ports"] = {"%i/tcp" % self.port: None}
 
-        create_kwargs.update(self.extra_create_kwargs)
+        create_kwargs.update(self._extra_create_kwargs_subs())
```
minrk, line 891: Presumably apply the same substitutions to host config

dockerspawner/dockerspawner.py
```diff
@@ -1173,3 +1172,22 @@ def _fmt(v):
                 v = v["bind"]
             binds[_fmt(k)] = {"bind": _fmt(v), "mode": m}
         return binds
+
+    def _extra_create_kwargs_subs(self):
```
minrk, line 1176: Rather than having a method return only extra_create_kwargs, this should probably be implemented as a generic walker of reformatting a potentially deep container:

- walk lists and dicts
- handle docker models objects? Not sure if this is feasible or necessary.
- where strings are seen (always check that it's a string) -> apply `_fmt(s)`

minrk (commented):
Sorry for commenting prematurely. There are a number of cases where this will raise errors if given the wrong types.

It would be good to add some unittests to verify with certain combinations of extra_create_kwargs and extra_host_config that the expected result is produced.

I'm happy to help with that if needed

devnull-mr:
Hi,

thanks for the review and fixes in #409 !
If it works for you I'm fine !

minrk:
Thanks @devnull-mr! closed in favor of #409

PR #373, opened 2020-06-16
Add mounts option for more advanced binds

rowleya:
This allows the user to specify mounts using a dict, which are then converted to docker.types.Mount objects when spawning.  This allows, for example, the use of propagation, which is useful when mounting a disk that is already mounted locally e.g. a fuse mount.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -15,6 +15,7 @@ import warnings
 import docker
 from docker.errors import APIError
 from docker.utils import kwargs_from_env
+from docker.types import Mount
 from tornado import gen, web
 
 from escapism import escape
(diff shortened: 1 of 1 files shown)


TimoRoth:
Hi!
I was about to write the exact same thing, and luckily saw your PR before starting.
I'd have one feature request: Could you add templating to the mounts, in the same way it's done for the volumes?
For the setup I'm planning I need the substitute the {username} into the mount source.

TimoRoth:
Finally figured out how to properly set this up. The documentation (from docker-py) on this is very poor.
The correct way to pass the equivalent of "--mount src=jupyter-storage-{username},dst=/home,type=volume,volume-opt=size=16G" is:

```
c.DockerSpawner.mounts.append({
    'source': 'jupyter-storage-{username}',
    'target': '/home/jovyan',
    'type': 'volume',
    'driver_config': {
        'Options': {
            'size': '16G'
        }
    }
})
```

Extremely non-obvious and not typo safe at all, but works as intended.

+1 for merging this PR from me

minrk:
This is great! Sorry for missing it. I merged the first commit in #406, so without the 'optional' argument. If you think deviating from the mounts spec to support optional mounts is important, a new PR would be welcome!

PR #374, opened 2020-06-16
Fix tests

GeorgianaElena:
- since `pytest-cov:2.10.0`, [pytest>=4.6 is required](https://github.com/pytest-dev/pytest-cov/blob/master/CHANGELOG.rst#2100-2020-06-12)
- `notebook>=6.1.*` errors if [Tornado < 5](https://github.com/jupyter/notebook/blob/e937fd79927e5806656a4229091d9814686626d2/docs/source/changelog.rst#610) (at the moment there is only a beta release for `notebook:6.1.0`, but once a stable version is release, the tests will fail again)
- the CI jobs were a bit re-organized

This is only a temporary fix, because a new release of the `notebook` will break the tests again.
I think a more permanent solution would be to test instead with the latest two JupyterHub versions (1.0 and 1.1.0). This means we should:
- switch from `pytest-tornado` to `pytest-asyncio`
- have a working jupyterhub/singleuser version for JupyterHub 1.1.0 (ref: https://github.com/jupyterhub/jupyterhub/issues/2911)

Update: Just realized that pinning pytest-cov to 2.9.0 would have probably been a easier fix :sweat_smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,22 @@
-sudo: required
 language: python
-python:
-  - 3.5
 env:
   global:
     - ASYNC_TEST_TIMEOUT=30
-  matrix:
-    - JUPYTERHUB=0.8 EXTRA_PIP='tornado<5'
-    - JUPYTERHUB=0.9
+
+# request additional services for the jobs to access
 services:
   - docker
+
 before_install:
   - docker swarm init
   - nvm install 10; nvm use 10
   - npm install -g configurable-http-proxy
   - pip install --upgrade -r dev-requirements.txt jupyterhub==${JUPYTERHUB:-0}.* ${EXTRA_PIP}
+
 install:
   - pip install -e .
   - pip freeze
+
 script:
   - pyflakes dockerspawner
   - |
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -41,7 +40,22 @@ script:
...
       env: TEST=internal-ssl
+    - name: python:3.5 + jupyterhub 0.8 + tornado < 5
+      python: 3.5
+      env: JUPYTERHUB=0.8 EXTRA_PIP='tornado<5'
+    - name: python:3.6 + jupyterhub 0.8 + tornado < 5
+      python: 3.6
+      env: JUPYTERHUB=0.8 EXTRA_PIP='tornado<5'
+    - name: python:3.7 + jupyterhub 0.9
+      python: 3.7
+      env: JUPYTERHUB=0.9
```
manics, line 56: Since Python 3.7 and 3.8 are new scenarios could you test the latest jupyterhub? Or is that going to break something?

manics:
One comment, but this seems to fix the tests.

If you want to go for your proposed solution (update to Jupyterhub 1.*) I think that's fine, this project hasn't had a release for ages so I expect any installations of the next release will be against a more recent JupyterHub.

.travis.yml
GeorgianaElena: I believe it is going to fail, because in JupyterHub 1.0, the tests are using `pytest-asyncio` instead of `pytest-tornado`. Dockerspawner uses some of the fixtures JH, so we'll have to switch here too before testing with latest JupyterHubs.

I started working on this [here](https://github.com/jupyterhub/dockerspawner/compare/master...GeorgianaElena:fix_tests) and managed to get it working for JH 1.0, but for 1.1.0 it's failing because of the issue with 1.1.0 docker tags of the `jupyterhub/singleuser` image.

manics: Thanks! Let's get this in and deal with the update later.

Issue #332, opened 2019-09-05
readthedocs


minrk:
We should have a readthedocs site with configurables populated by autoconfig (or whatever it's called. It's in the jupyterhub repo)

willingc:
Check ipython for autoconfig extension. Jim's YarnSpawner docs are nice.

Create a package on PyPI for autoconfig extension.

PR #375, opened 2020-06-17
Add docs

GeorgianaElena:
- [x] Autodoc traits
- [x] Readthedocs config
- [x] Readthedocs site:
   https://readthedocs.org/projects/jupyterhub-dockerspawner/
- [x] Build documentation previews for PRs with CircleCi
    Docs preview here: https://18-236708530-gh.circle-artifacts.com/0/html/index.html

Closes https://github.com/jupyterhub/dockerspawner/issues/332


GeorgianaElena:
Thanks for reviewing this @choldgraf :hugs: 

> Maybe we should create a JupyterHub/Dockerspawner logo like we've done elsewhere (see https://docs.google.com/presentation/d/11Ag7oj8zaRbbTSLFTMYEAP-7WWCLFJRkKaEjjjE-nSA/edit#slide=id.p for inspiration)

Great idea! :heart:  What do you think about this one?
jh-dockerspawner-logo
 
> Since each page of this site does not have sub-pages, perhaps we could use the sphinx-book-theme instead of the pydata-sphinx-theme? They are similar but the SBT is better for simpler sites, I think (https://sphinx-book-theme.readthedocs.io/)

Sure! I think we should try it out :+1: 

> In general, I think this is a great idea to use docs instead of a gigantic README file. Did you change the content much, or is mostly just structural?

Yes, the docs content is mostly just copy-paste from readme.

choldgraf:
the logo looks great to me!

choldgraf:
@GeorgianaElena looking good! One comment on the logo: you've got a fairly "square" amount of white-space around the logo:

image

What I'd recommend is cropping it so that the logo is wide and short, that way it will take up a bit less space in the final docs. What do you think?

Other than that, anything else you wanna change before merging?

GeorgianaElena:
Thanks Chris! I've just cropped the logo :+1: 
> Other than that, anything else you wanna change before merging?

No, I think it should be ok for now.

docs/source/conf.py
```diff
@@ -0,0 +1,75 @@
...
+
+# -- Options for HTML output -------------------------------------------------
+
+# The theme to use for HTML and HTML Help pages.  See the documentation for
+# a list of builtin themes.
+#
+html_theme = 'sphinx_book_theme'
+html_logo = '_static/images/logo/logo.png'
+html_favicon = '_static/images/logo/favicon.ico'
+
```
choldgraf, line 70:
```suggestion
html_title = ''

```

choldgraf:
just fixed up one thing to remove the extra title, and I think this is good to go! thanks @GeorgianaElena !!!

PR #377, opened 2020-06-23
Add release instructions and Travis deploy

GeorgianaElena:
Adds release instructions and a travis deploy job.

@consideRatio, do you think you can help with the `PYPI_PASSWORD` env var setup here too? :heart:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,7 +41,9 @@ script:
     fi
 
 jobs:
+  fast_finish: true
   include:
+    # Default stage: test
     - name: internal-ssl test
       python: 3.6
       env: TEST=internal-ssl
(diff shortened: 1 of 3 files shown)


GeorgianaElena:
I think we need Min's help to get the the PyPI user `jupyterhub-bot` added as a PyPI package maintainer  (https://pypi.org/project/dockerspawner/) and set the  `PyPI API token` for this repo.

@minrk, do you think you help out with this? :heart: 

It would be really nice to have a new release for `dockerspawner` as last one was more that a year ago.
What do you think?

consideRatio:
I think you are right @GeorgianaElena, sorry for slow followup! Following that, a visit to https://travis-ci.org/github/jupyterhub/dockerspawner/settings where a PYPI_PASSWORD env var is set with the deployment token created for the jupyterhub-bot for this repo.

minrk:
Great!

PR #379, opened 2020-06-24
Test with latest jh

GeorgianaElena:
- Updates the tests to work with latest JupyterHub versions: 1.0 and 1.1
- Use `pytest-asyncio` instead of `pytest-tornado`.

Note: `jupyterhub/singleuser:1.1` doesn't exist due to an issue with docker tags so the tests for JH 1.1 fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,12 +53,12 @@ jobs:
     - name: python:3.6 + jupyterhub 1.0
       python: 3.6
       env: JUPYTERHUB=1.0
-    - name: python:3.7 + jupyterhub 1.2.0dev
+    - name: python:3.7 + jupyterhub 1.1.0
       python: 3.7
-      env: JUPYTERHUB=1.2.0dev
-    - name: python:3.8 + jupyterhub 1.2.0dev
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -21,13 +21,13 @@ script:
   - pyflakes dockerspawner
   - |
     # pre-pull images to avoid long wait times in tests
-    for v in 0.8 0.9; do
```
manics, line 24: Presumably the intention is to restore the for loop in future to check both 1.0 and 1.1 (as opposed to only 1.1)? If so I think keeping the for loop with just one item `for v in 1.0; do` will make it slightly easier for whoever fixes this in future.

minrk: I just pushed the 1.1 image manually, so it should work now, and opened https://github.com/jupyterhub/jupyterhub/pull/3118 to fix it in the future

GeorgianaElena: Yaaay. Thank you :heart: 
I've just dropped the last commit that removed the tests for 1.1 and it seems that everything is green now.

minrk:
Yay, thank you!

PR #380, opened 2020-06-24
Typo fix in the docs

jtpio:
Really nice work on the docs!

Spotted a minor typo while going through them.


jtpio:
In a separate issue / PR, we could also extend the "How to build your own Docker image" section to mention `repo2docker` as a way to build Docker images.

PR #381, opened 2020-06-25
Rename DockerSpawner.whitelist to allowed

GeorgianaElena:
cf https://github.com/jupyterhub/team-compass/issues/299
https://github.com/jupyterhub/jupyterhub/pull/3090

- [x] DockerSpawner.image_whitelist -> DockerSpawner.allowed_images
- [x] DockerSpawner.check_image_whitelist -> DockerSpawner.check_allowed
- [x] deprecate traits, config
- [x] test deprecated names still work
- [x] test deprecated config
- [x] update docs (no occurrences in docs)

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -1,7 +1,6 @@
 """
 A Spawner for JupyterHub that runs each user's server in a separate docker container
 """
-
 from concurrent.futures import ThreadPoolExecutor
 from io import BytesIO
 import os
@@ -15,6 +14,7 @@ import warnings
 import docker
 from docker.errors import APIError
 from docker.utils import kwargs_from_env
+from functools import partial
 from tornado import gen, web
 
 from escapism import escape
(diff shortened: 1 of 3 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -54,8 +54,72 @@ def validate(self, obj, value):
 class DockerSpawner(Spawner):
     """A Spawner for JupyterHub that runs each user's server in a separate docker container"""
 
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+        self._init_deprecated_methods()
+
+    def _init_deprecated_methods(self):
```
minrk, line 61: Unlike the Authenticators, where subclassing and overriding methods is a standard approach, I don't think subclassing and overriding these methods is something we need to keep on DockerSpawner.

dockerspawner/dockerspawner.py
```diff
@@ -54,8 +54,72 @@ def validate(self, obj, value):
...
+                def _new_calls_old(old_name, *args, **kwargs):
+                    return getattr(self, old_name)(*args, **kwargs)
+
+                setattr(self, new_name, partial(_new_calls_old, old_name))
+                break
+
     _executor = None
 
+    _deprecated_aliases = {
+        "container_ip": ("host_ip", "0.9.*"),
```
minrk, line 93: Great job consolidating these!

dockerspawner/dockerspawner.py
GeorgianaElena: It makes sense. Thanks for explaining this, Min!

minrk:
Excellent, thank you!

PR #386, opened 2020-07-17
avoid name collisions when using named servers

minrk:
we haven't made a release yet with #315 (merged last month), but the default template added there allows collisions between certain usernames and named servers because `-` is considered a safe character and not escaped.

This changes the default template to use `__`, the doubled escape character (same as on KubeSpawner, which uses `-` as the escape character), which cannot occur in an escaped username or servername.

This also ensures the escaped form of servername is used in the template, matching handling of usernames.

Additionally, this persists object_name in state, to avoid changes in the template (as is done here) from causing a disconnect between objects across configuration changes.
That is, changes in the template will only affect newly created objects, not ones that already exist.

/cc @danielballan

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ before_install:
   - docker swarm init
   - nvm install 10; nvm use 10
   - npm install -g configurable-http-proxy
+  - pip install --upgrade setuptools pip
   - pip install --upgrade -r dev-requirements.txt jupyterhub==${JUPYTERHUB:-0}.*
 
 install:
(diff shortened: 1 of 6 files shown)


danielballan:
Good catch! Thanks for cleaning up after me. :- D

minrk:
Tests are failing because conftests imports some fixtures from jupyterhub that are new in 1.0. Need to work around that...

manics:
@GeorgianaElena already started work on updating the tests for JupyterHub 1.0 https://github.com/jupyterhub/dockerspawner/pull/379

GeorgianaElena:
I didn't know what to do about the `jupyterhub/singleuser:1.1` image :confused: 
I'll rebase that PR and maybe test with JupyterHub 1.0 only? What do you think?

GeorgianaElena:
Thanks, Min :heart: !

PR #387, opened 2020-07-20
stop building docs on circleci

minrk:
we build docs on RTD now


minrk:
I had to manually login to circle to stop building. I expected it to just stop when the file was gone, like travis, but instead it raised an error about the missing file.

Issue #330, opened 2019-08-27
Timeout potentially due to docker inspect_container prior to creation


gdbassett:
I have a jupyterhub environment using swarmspawner.  It works fine in my local testing, but when deployed to my rather old and slow production environment, it fails with the attached logs.  However, when I look at the services, the service is up and running successfully.  If I trace it back, it seems to go https://github.com/jupyterhub/dockerspawner/blob/9d4a35995d2c2dd992e070cc7ad260123308b606/dockerspawner/swarmspawner.py#L252 through get_task() to https://github.com/jupyterhub/dockerspawner/blob/9d4a35995d2c2dd992e070cc7ad260123308b606/dockerspawner/dockerspawner.py#L781 which calls inspect_service() from https://docker-py.readthedocs.io/en/stable/api.html#module-docker.api.service.  If I manually run inspect_service() and tasks() I get back a running service.

However, inspecting during startup, I get this:
```
>>> for task in client.tasks(filters={"service": "jupyter-gdb"}):
...    print(task['Status']['State'])
... 
rejected
rejected
>>> for task in client.tasks(filters={"service": "jupyter-gdb"}):
...    print(task['Status']['State'])
... 
rejected
rejected
rejected
>>> for task in client.tasks(filters={"service": "jupyter-gdb"}):
...    print(task['Status']['State'])
... 
rejected
rejected
rejected
rejected
>>> for task in client.tasks(filters={"service": "jupyter-gdb"}):
...    print(task['Status']['State'])
... 
rejected
rejected
rejected
rejected
rejected
>>> for task in client.tasks(filters={"service": "jupyter-gdb"}):
...    print(task['Status']['State'])
... 
rejected
rejected
rejected
ready
rejected
>>> for task in client.tasks(filters={"service": "jupyter-gdb"}):
...    print(task['Status']['State'])
... 
rejected
rejected
rejected
running
rejected
```
For some reason, it appears as if the task gets rejected for a while before running and the swarmspawner picks this up as a failure.  I wonder if it has to do https://github.com/jupyterhub/dockerspawner/blob/9d4a35995d2c2dd992e070cc7ad260123308b606/dockerspawner/swarmspawner.py#L257 checking 'State' instead of 'Message', as Message is 'preparing' when State is 'rejected':

```
  'Status': {'ContainerStatus': {},
             'Err': 'No such image: '
                    '<image name and hash>',
             'Message': 'preparing',
             'PortStatus': {},
             'State': 'rejected',
[...]

PR #396, opened 2020-11-06
Make sure that create_object() creates the service task

girgink:
Occasionally single-user task is not created immediately when `create_service` call yields in the `create_object()` method. Therefore, it is necessary to wait until it is created. Wait loop added to the end of `create_object()` method guarantees this. This problem is linked to [Issue 330](https://github.com/jupyterhub/dockerspawner/issues/330) and I think the update solves the last unsolved (e.g. "we are still seeing") part, which seems to occur more frequently if swarm includes nodes with different architectures (e.g. Arm64v8 and AMD64) (it is also my case).

The change also includes the patch by @gdbassett [mentioned in Issue 330](https://github.com/jupyterhub/dockerspawner/issues/330#issuecomment-525489513) to property get the task created by the service and check its states. `ready` state is not listed in the official documents ([Swarm task states](https://docs.docker.com/engine/swarm/how-swarm-mode-works/swarm-task-states/)), but I saw it in the debug logs (Docker v 19.03.8). If not added, these states break the loop and results in RuntimeError afterwards.

### DIFF
--- a/dockerspawner/swarmspawner.py
+++ b/dockerspawner/swarmspawner.py
@@ -151,7 +151,12 @@ class SwarmSpawner(DockerSpawner):
                 filters={"service": self.service_name, "desired-state": "running"},
             )
             if len(tasks) == 0:
-                return None
+                tasks = yield self.docker(
+                    "tasks",
+                    filters={"service": self.service_name},
+                )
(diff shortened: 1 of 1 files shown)


minrk:
Thanks!

Issue #285, opened 2019-02-15
SwarmSpawner.image_whitelist spawn only default image


zi-dan:
Hi Guys,

I try to use image_whitelist in my SwarmSpawner with guidance of issue #219  but always a default image is choosen. It looks like a image_whitelist options are skipped and always values of SwarmSpawner.image is taken from configuration. 
My jupyter_config has following parameters:

```
c.JupyterHub.spawner_class = 'dockerspawner.SwarmSpawner' 
c.SwarmSpawner.image = 'jupyter/minimal-notebook:77e10160c7ef' 
c.SwarmSpawner.image_whitelist = { 
    'Jupyteo All-In-One': 'jupytepide/user-spawn-notebook:dev', 
    'Jupyteo EO Processing': 'jupytepide/eodata-notebook:1.3.6' 
}
```
With this config list with option is displayed but choosing any of the item loads jupyter/minimal-notebook.

Did I miss something?

minrk:
were you using dockerspawner master? This isn't in a release yet, though I'm trying to make the 0.11 release today.

zi-dan:
Yes, I use a master branch. So I will wait for update and new release.

mathematicalmichael:
hi, I cloned off master (`pip install git+https://github.com/jupyterhub/dockerspawner.git`) and switched to  `allowed_images`, but I'm still only getting default spawns. Why might this be?

efagerberg:
I have not tried master but I tried 0.11.1 with the allowed_images option and also only get the default image with no form. In my case I do not specify `c.DockerSpawner.image` at all so it seems like `allowed_images` is being looked and just taking the first image.

GeorgianaElena:
I didn't manage to reproduce this with:
- `jupyterhub 1.1.0`
- `dockerspawner 0.12.0.dev0 (master)`

and @zi-dan's config:
```python
c.JupyterHub.spawner_class = 'dockerspawner.DockerSpawner'
c.DockerSpawner.allowed_images = {
    'Jupyteo All-In-One': 'jupytepide/user-spawn-notebook:dev',
    'Jupyteo EO Processing': 'jupytepide/eodata-notebook:1.3.6'
}
```
:confused: 
![bug-not-reproducing](https://user-images.githubusercontent.com/7579677/89995736-8e5cbe00-dc92-11ea-8f7b-2115639dca3e.gif)

efagerberg:
[...]

PR #399, opened 2020-11-12
Fix issue #285. Now can spawn more than one image on jupyterhubs 1.2.…

slemonide:
…0 and up.

Seems to fix the issue.


minrk:
The method name should not be relevant, so I don't think this should change any behavior. Maybe the difference you are seeing is between the latest stable release and master?

minrk:
Thanks for the PR! I'm going to close this one, since the method name has no effect and #285 is closed.

PR #407, opened 2021-02-25
Switch CI to GitHub actions

minrk:
- remove travis CI
- adopt pre-commit config from other projects, runs big auto-format commit
- bump required Python to 3.6
- bump required jupyterhub to 1.0
- fix a small bug in a test


minrk:
Since this PR adds the actions, they won't run until the PR is merged. Results can be seen [on my fork](https://github.com/minrk/dockerspawner/actions), where everything is green.

.github/workflows/test.yml
```diff
@@ -0,0 +1,111 @@
...
+
+      # preserve pip cache to speed up installation
+      - name: Cache pip
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/pip
+          # Look to see if there is a cache hit for the corresponding requirements file
+          key: ${{ runner.os }}-pip-${{ hashFiles('*requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
```
GeorgianaElena, line 66: I have bumped over this issue https://github.com/pypa/pip/issues/4537 when trying to make sense on why the TLJH unit tests [are failing](https://app.circleci.com/pipelines/github/jupyterhub/the-littlest-jupyterhub/219/workflows/5088d3ab-cc2e-4c60-909f-a72e15e9c37d/jobs/2740).
A possible solution I had in mind was to only keep the most restrictive key and don't have a restore one. Basically if anything changes in the requirement.txt, then just install everything.

I can't say if it's the case here too because I still don't think I fully understand the underlying issue :( But I thought it's worth knowing about this.

GeorgianaElena (approved):
:heart_eyes: This is awesome! Thank you @minrk!

.github/workflows/test.yml
```diff
@@ -0,0 +1,112 @@
...
+#
+name: Tests
+
+on:
+  pull_request:
+  push:
+
+defaults:
+  run:
+    shell: bash
```
consideRatio, line 12: This is a legacy of me thinking the default shell was `sh` but it is actually `bash`! Can be removed.

.github/workflows/test.yml
```diff
@@ -0,0 +1,112 @@
+# This is a GitHub workflow defining a set of jobs with a set of steps.
+# ref: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions
+#
+name: Tests
+
+on:
+  pull_request:
+  push:
```
consideRatio, line 8: I like having `workflow_dispatch:` here as well, it provides maintainers with a button to trigger the workflow manually. This can be useful to verify how tests work on current master for example. It is relevant whenever you observe a failure in a PR and get confused if it is because of the PR or the master.

consideRatio (approved):
Nice this LGTM!

I suggest two non-critical changes:
- [x] Add `workflow_dispatch:` trigger, provides a UI button for maintainers to trigger a workflow.
- [ ] Remove defaults.run.shell introduced by me misunderstanding what the default environment were - it is redundant to specify bash as I understand it now.

minrk:
Thanks! Applied all your suggestions.

consideRatio:
:tada: !!!

consideRatio:
When we migrate we always forget the readme badge ;D

https://github.com/jupyterhub/dockerspawner/pull/408

PR #411, opened 2021-02-26
add 'skip' option for pull_policy

### LINKED ISSUE #394: Swarmspawner tries to always pull image on manager
Even if swarmspawner is used https://github.com/jupyterhub/dockerspawner/blob/b6fccbf8c40ff22486dfdd388bf577ef3696ed47/dockerspawner/dockerspawner.py#L966 (default) tries to pull the notebook image on the manager. 

In our use case (which I think is not so uncommon) we do not want to run notebooks on the manager node and have disabled that by a deploy option. It is therefore un-logical that the code checks for the existence of the the notebook image on the manager. Setting pull_policy to 'never' does not fix the issue as then it aborts as soon as it does not find the image on the manager (even if it exists on the worker nodes).

Ideally swarmspawner should check/pull the specified image on the node that is assigned the task but I am unsure if docker swarm allows such checks/actions (seams like service create only check if the tags exist on the node not if it has been updated that the source). We have already had a bug where the image on the worker nodes where not updated (as we reused the :latest tag). However a more proper use of tagging resolved those issues of course.

minrk:
skips unhelpful pull by default on swarm because the pull won't usually happen on the same node as the container anyway

closes #394

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -298,7 +298,7 @@ class DockerSpawner(Spawner):
         return options
 
     pull_policy = CaselessStrEnum(
-        ["always", "ifnotpresent", "never"],
+        ["always", "ifnotpresent", "never", "skip"],
         default_value="ifnotpresent",
         config=True,
         help="""The policy for pulling the user docker image.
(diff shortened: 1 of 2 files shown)


yuvipanda:
I like!

PR #414, opened 2021-03-11
Make escaping DNS-safe

### LINKED ISSUE #205: Usernames are incorrectly sanitized
Dockerspawner provides the  `escaped_username` property, which is internally used to populate `username` field in the `container_name_template`.

If I understand correctly this is because docker containers may not start with an underscore. I note that when trying to start a docker container with the name `_`, docker informs me of the following:

```
docker: Error response from daemon: Invalid container name (_), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.
```

If  `container_name_template` is set to `{username}`, then we must ensure that usernames do indeed conform to the above regex. This does not, however, appear to be the case. 

Consider the following:

```
from types import SimpleNamespace
import dockerspawner

ds = dockerspawner.DockerSpawner()
ds.user = SimpleNamespace(name='_')  # mock
print(ds.escaped_name)  # prints '_5F' 

ds = dockerspawner.DockerSpawner()
ds.user = SimpleNamespace(name='hello_')  # mock
print(ds.escaped_name) # >>> 'hello_5F' 
```

It appears that `_` is being used as a marker for escaped characters (as `%` is in URLs) however, as previously noted, this is not safe.

### LINKED ISSUE #209: c.DockerSpawner.volumes invalid characters for a local volume
This is related to how {username} gets defined when an OAuth user comes in to create a docker volume.

My current config:
`notebook_dir` = '/home/jovyan/work'
c.DockerSpawner.notebook_dir = notebook_dir
c.DockerSpawner.volumes = { 'jupyterhub-user-{username}': notebook_dir `}`

The resultant error when a server is started:

 400 Client Error: Bad Request ("create jupyterhub-user-X.Y.Z@gmail.com: "jupyterhub-user-X.Y.Z@gmail.com" includes invalid characters for a local volume name, only "[a-zA-Z0-9][a-zA-Z0-9_.-]"

I though i might be able to populate the c.DockerSpawner.volumes dictionary with regex to match everything before '@', but so far been unsuccessful because {username} is populated at runtime and not at jupyterhub init

c.DockerSpawner.volumes = { re.search(r'.*(?=@)', '{username}') : notebook_dir }

Any thoughts as to how I can accomplish this?

### LINKED ISSUE #262: SwarmSpawner unable to contact notebook container when container name has upper case characters

### LINKED ISSUE #305: SwarmSpawner unable to reach containers with escaped characters
I am attempting to get a JupyterHub deploy set up and working with SwarmSpawner. It works... except for users who have a special character in their username.

I believe this is an issue in how SwarmSpawner is handling special characters.

When it encounters special characters it escapes them using an underscore. For example `tabor@cloud.ucsd.edu` becomes `tabor_40cloud_2Eucsd_2Eedu`. This is used in the name of the Docker service, which in turn is used in the hostname.

I have verified that the Docker service is being started. But then when JupyterHub tries to contact the single-user server at `http://jupyter-tabor_40cloud_2Eucsd_2Eedu:8888/user/tabor@cloud.ucsd.edu/` (or whatever the username with special characters may be), it times out with a `socket.gaierror: [Errno -2] "Name or service not known"` error.

I suspect that this is because underscores aren't a valid hostname character.

### LINKED ISSUE #326: Container name escaping doesn't align with latest Docker rules
(Possibly related: https://github.com/jupyterhub/dockerspawner/issues/205, https://github.com/jupyterhub/dockerspawner/issues/305)

## Some history + the problem

1. In https://github.com/jupyterhub/dockerspawner/pull/40, `dockerspawner` gained the ability to escape "special characters" in container names.
2. ^ This commit was built in reaction to this change to Docker to require that container names match the regex `^/?[a-zA-Z0-9_-]+$`: https://github.com/moby/moby/commit/f63cdf0260cf6287d28a589a79d3f947def6a569

3. Docker has since changed their container naming rules a couple times:
   - Allowing periods: https://github.com/moby/moby/commit/1940015824f
   - Requiring all container names to start with `a-zA-Z0-9`: https://github.com/moby/moby/commit/aa2af817bee123827f4a857555db8fea5e01e7d7

Currently, Docker requires container names to match `^[a-zA-Z0-9][a-zA-Z0-9_.-]$`: https://github.com/moby/moby/blob/928381b2215c9608692e663e6ddfa9e609bc2d6f/daemon/names/names.go#L5-L6, which is is different from [`dockerspawner`'s current escaping logic](https://github.com/jupyterhub/dockerspawner/blob/9d4a35995d2c2dd992e070cc7ad260123308b606/dockerspawner/dockerspawner.py#L672-L678) in two important ways:
  1. `dockerspawner` unnecessarily treats `.` as unsafe. You could say this isn't exactly a bug, but it is a bit weird.
  2. `_`, `.`, and `-` are only acceptable container names if they are *not* the first character in the container name. Right now, if someone's username starts with a special character, `dockerspawner` will escape that special character and generate an escaped container name that starts with a `_`. For example:
     ```
     $ pip freeze | grep -i dockerspawner
     dockerspawner==0.11.1
 
     $ python
     Python 3.7.4 (default, Jul 16 2019, 07:12:58) 
     >>> import dockerspawner
     >>> ds = dockerspawner.DockerSpawner()
     >>> ds._escape("$Money Name$")
     '_24Money_20Name_24'
     ```
     Proof that this container name won't work with docker:
 
     ```
     $ docker run --name "_24Money_20Name_24" ubuntu
     docker: Error response from daemon: Invalid container name (_24Money_20Name_24), only [a-zA-Z0-9][a-zA-Z0-9_.-] are allowed.
     See 'docker run --help'.
     ```

## Proposals

I think the right thing to do would be for `dockerspawnner` to add `.` to the `_docker_safe_chars` list (to match Docker's current behavior), and also update the escaping logic to never generate `_` as the first character in a name. Some ideas:

1. Change [`_docker_escape_char`](https://github.com/jupyterhub/dockerspawner/blob/9d4a35995d2c2dd992e070cc7ad260123308b606/dockerspawner/dockerspawner.py#L525) from `_` to something that *is* an allowed first character container name.
   - This is pretty simple, but all the choices we could make (`[a-zA-Z0-9]`) are pretty unfortunate.
2. Prefix all container names with something known to be safe. Something like "dockerspawner-"?
   - This is pretty simple, and probably less weird than 1) above.
3. Handle the first character of the container name specially in some weird way. Maybe say "we'll use `0` as an escape character for the *first* character of the name, and then use `_` as the escape character for the rest of the name.
   - Cons: This is pretty weird. The decoding logic would also be weird, but I'm not sure that `dockerspawner` actually ever decodes the container names it produces.
   - Pros: It's the most similar of all these options to the current behavior, but still obeys all of Docker's current container naming rules.

Happy to send in a PR if someone wants to give guidance on which of 1, 2, 3 above sound good.

minrk:
While containers allow more characters, networking and the like are more strict.

To ensure things like DNS work (required for swarm, internal ssl), use a stricter DNS-safe subset for escaping.

c.DockerSpawner.escape is now a configurable and can be overridden to preserve old behavior and/or use other schemes.

closes #205
closes #326
closes #305
closes #342
closes #262
closes #209

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -341,13 +341,13 @@ class DockerSpawner(Spawner):
             It is important to include {servername} if JupyterHub's "named
             servers" are enabled (JupyterHub.allow_named_servers = True).
             If the server is named, the default name_template is
-            "{prefix}-{username}__{servername}". If it is unnamed, the default
+            "{prefix}-{username}--{servername}". If it is unnamed, the default
             name_template is "{prefix}-{username}".
 
             Note: when using named servers,
             it is important that the separator between {username} and {servername}
             is not a character that can occur in an escaped {username},
-            and also not the single escape character '_'.
+            and also not the single escape character '-'.
             """
         ),
     )
@@ -355,7 +355,7 @@ class DockerSpawner(Spawner):
     @default('name_template')
     def _default_name_template(self):
         if self.name:
-            return "{prefix}-{username}__{servername}"
+            return "{prefix}-{username}--{servername}"
         else:
             return "{prefix}-{username}"
 
@@ -571,8 +571,60 @@ class DockerSpawner(Spawner):
         config=True, help="Additional args to create_host_config for container create"
     )
 
-    _docker_safe_chars = set(string.ascii_letters + string.digits + "-")
-    _docker_escape_char = "_"
+    escape = Any(
+        help="""Override escaping with any callable of the form escape(str)->str
+
+        This is used to ensure docker-safe container names, etc.
+
+        The default escaping should ensure safety and validity,
+        but can produce cumbersome strings in cases.
+
+        Set c.DockerSpawner.escape = 'legacy' to preserve the earlier, unsafe behavior
+        if it worked for you.
+
+        .. versionadded:: 0.12
+
+        .. versionchanged:: 0.12
+            Escaping has changed in 0.12 to ensure safety,
+            but existing deployments will get different container and volume names.
+        """,
+        config=True,
+    )
+
+    @default("escape")
+    def _escape_default(self):
+        return self._escape
+
+    @validate("escape")
+    def _validate_escape(self, proposal):
+        escape = proposal.value
+        if escape == "legacy":
+            return self._legacy_escape
+        if not callable(escape):
+            raise ValueError("DockerSpawner.escape must be callable, got %r" % escape)
+        return escape
+
+    @staticmethod
+    def _escape(text):
+        # Make sure a substring matches the restrictions for DNS labels
+        # Note: '-' cannot be in safe_chars, as it is being used as escape character
+        # any '-' must be escaped to '-2d' to avoid collisions
+        safe_chars = set(string.ascii_lowercase + string.digits)
+        return escape(text, safe_chars, escape_char='-').lower()
+
+    @staticmethod
+    def _legacy_escape(text):
+        """Legacy implementation of escape
+
+        Select with config c.DockerSpawner.escape = 'legacy'
+
+        Unsafe and doesn't work in all cases,
+        but allows opt-in to backward compatibility for an upgrading deployment.
+
+        Do not use for new deployments.
+        """
+        safe_chars = set(string.ascii_letters + string.digits + "-")
+        return escape(text, safe_chars, escape_char='_')
 
     hub_ip_connect = Unicode(
         config=True,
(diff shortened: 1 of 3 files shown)


consideRatio (approved):
This LGTM!

PR #415, opened 2021-03-11
Use default cmd=None to indicate using the image command

### LINKED ISSUE #317: Using _user_set_cmd to override cmd in custom spawner
When defining a custom spawner that inherits from `DockerSpawner`, it looks like the `_user_set_cmd` flag needs to be set manually so `cmd` can be picked up. For example:

```python
class CustomSpawner(DockerSpawner):
    _user_set_cmd = True
    cmd = ['jupyterhub-singleuser']

...

c.JupyterHub.spawner_class = CustomSpawner
```

Without setting `_user_set_cmd` manually the `cmd` would default to the `CMD` from the Docker image.

Maybe there is a better way to make sure the custom command is picked up without using `_user_set_cmd`?

minrk:
instead of an observer, which isn't kind to subclasses

closes #317

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -111,14 +111,10 @@ class DockerSpawner(Spawner):
             cls._client = client
         return cls._client
 
-    # notice when user has set the command
-    # default command is that of the container,
-    # but user can override it via config
-    _user_set_cmd = False
-
-    @observe("cmd")
-    def _cmd_changed(self, change):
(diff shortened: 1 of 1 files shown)


consideRatio (approved):
Ah the Spawner base class set `cmd` to `[jupyterhub-singleuser]`, but DockerSpawner wanted another default and accomplished that not by using a `@default` decorator but by observing if `cmd` was changed.

This improves readability A LOT and shouldn't be breaking in any way.

PR #420, opened 2021-03-22
Allow to specify a callable for `mem_limit`

zeehio:
Similarly to the `allowed_images` setting, it would be very convenient to be able to set the `mem_limit` to a callable, to have more flexibility to enforce memory limits.

This feature would for instance allow easily define user-specific memory limits in `jupyterhub_config.py`:

```python
def per_user_mem_limit(spawner):
    username = spawner.user.name
    ram_limits = {}
    try:
        with open("/srv/jupyterhub/ram_limits.json") as f:
            ram_limits = json.load(f)
    except:
        pass
    return ram_limits.get(username, "32G")
c.SystemUserSpawner.mem_limit = per_user_mem_limit
```

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -21,6 +21,8 @@ from docker.types import Mount
 from docker.utils import kwargs_from_env
 from escapism import escape
 from jupyterhub.spawner import Spawner
+from jupyterhub.traitlets import ByteSpecification
+from jupyterhub.traitlets import Callable
 from tornado import web
 from traitlets import Any
 from traitlets import Bool
@@ -54,6 +56,15 @@ _jupyterhub_xy = "%i.%i" % (jupyterhub.version_info[:2])
 
 class DockerSpawner(Spawner):
     """A Spawner for JupyterHub that runs each user's server in a separate docker container"""
+
+    def _eval_if_callable(self, x):
+        """Evaluate x if it is callable
+
+        Or return x otherwise
+        """
+        if callable(x):
+            return x(self)
+        return x
 
     _executor = None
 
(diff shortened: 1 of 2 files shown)


minrk:
👍 to the change. Would you like to make it a bit more systematic so more options can be callable?

I think we'll also need to modify the traitlet to allow it to accept a callable:

```python
from traitlets import Union
from jupyterhub.traitlets import ByteSpecification, Callable

...
mem_limit = Union([ByteSpecification(), Callable()], help="memory limit...").tag(config=True)
```

A unittest that sets this to be a callable should reveal that, I think.

zeehio:
I'll do some testing and make sure that my patch works as expected.

Thanks for the quick feedback

minrk (commented):
Awesome!

A parametrized unittest should be all that's left. Something like:

```python
@pytest.mark.parametrize(
    "mem_limit, expected",
    [
        ("1G", 1024 ** 3),
        (1_000_000, 1_000_000),
        (lambda spawner: "2G", 2 * 1024 ** 3),
        (lambda spawner: 1_000_000, 1_000_000),
    ],
)
def test_mem_limit(mem_limit, expected):
    s = DockerSpawner(mem_limit=mem_limit)
    assert s.mem_limit == expected
```

zeehio:
The unit test is failing on python3.6 & jupyterhub 1.0:

```

mem_limit = <function <lambda> at 0x7f3ad792f730>, expected = 2147483648

    @pytest.mark.parametrize(
        "mem_limit, expected",
        [
            ("1G", 1024 ** 3),
            (1_000_000, 1_000_000),
            (lambda spawner: "2G", 2 * 1024 ** 3),
            (lambda spawner: 1_000_000, 1_000_000),
        ],
    )
    def test_mem_limit(mem_limit, expected):
        s = DockerSpawner(mem_limit=mem_limit)
>       assert s.mem_limit == expected
E       AssertionError: assert '2G' == 2147483648
E         +'2G'
E         -2147483648

tests/test_dockerspawner.py:174: AssertionError
```

Was the casting somehow different on older jupyterhub versions?

minrk:
> Was the casting somehow different on older jupyterhub versions?

Must have been. Feel free to skip jupyterhub < 1.1 for that test.

minrk:
Thanks for this! Ignoring the internal-ssl failure, which is intermittent and not related to this one.

PR #412, opened 2021-02-26
add SystemUserSpawner.run_as_root

minrk:
when True, defer to image handing of $NB_UID and $NB_GID

this was previously a hard-to-opt-out default, but I think should be opt-in because images that do not support this will run all uses as root by default.

Downside of off-by-default: it's an extra flag needed to use docker-stacks to their full potential (they should still work without it, though, but usernames will not match, etc.)

Upside: userid is always right by default, no matter what image you use, no longer launch images as root that don't need it

closes #311
closes #283

PR #421, opened 2021-03-24
Changelog for 12.0

### LINKED ISSUE #366: Release a new version?
The last release was 12 months ago, with some bugfixes landing on master in the meantime (in particular #334 affects my deployment). Is there a plan to make a new (possibly bugfix) release?

minrk:
- [x] waiting for #412 at least

Highlights the breaking changes in run_as_root and container-name escaping.

Next version drops the leading 0, so planning to release 12.0 instead of 0.12.

closes #366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,117 @@ command line for details.
 
 ## [Unreleased]
 
+## 12
+
+### [12.0] 2021-03
+
+This is a big release!
+
+Several bugs have been fixed, especially in SwarmSpawner, and more configuration options added.
+
+#### New escaping scheme
+
+In particular, the biggest backward-incompatible change to highlight
+is the container (and volume) name escaping scheme now produces DNS-safe results, which matches the behavior of kubespawner.
+This is a stricter subset of characters than docker containers strictly require,
+but many features don't work right without it.
+The result is for certain user names and/or server names, their container and/or volume names will change.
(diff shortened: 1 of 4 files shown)


consideRatio:
@minrk is the practice of promoting the minor version to be the major version an established thing to do in situations when about to start using the major version number?

GeorgianaElena (approved):
LGTM :rocket:

(going from `0.12` to `12.0` is new for me too :astonished:)

minrk:
> If you can confirm that you really intended to release 12.0 rather than 0.12 or 1.0, then go for a merge!

It is intentional and I have done it before (on pyzmq), but I don't think it's common. I'm open to objections!

The idea is to get around the stress of "are we ready for 1.0?!" by skipping it entirely. 1.0 carries a lot of weight, but isn't really meaningful in practice. There is *lots* of software out there with a leading 0 that is mature and used widely in production, which generally means it shouldn't have a leading 0 anymore. But folks often worry too much about "what does 1.0 mean?" or adding a bunch of blockers such as "We need X and Y that we've been talking about for years before we can call it 1.0" so they end up sticking with 0.x for much longer than the maturity of the package suggests. There is even a perverse factor that the longer you wait to make 1.0, the *more* work it is to make a release that feels big enough to call it 1.0, when the opposite should be true. This strategy of "just drop the leading 0" is a way of communicating: **we are ready for general use, but this is also just the next release, not a special milestone**.

consideRatio:
@minrk well thought out! :) Sounds good to me!

Issue #432, opened 2021-07-14
CI has stopped


minrk:
### Bug description

CI has stopped passing, compare [last successful run](https://github.com/jupyterhub/dockerspawner/actions/runs/689633780) and the same commit [fails today](https://github.com/jupyterhub/dockerspawner/actions/runs/1029881955)

Since GHA aggressively purges test logs, it's hard to compare the working and failing test environments.

When I run the tests in a fresh env (on a mac), everything passes, which makes things hard to debug.

#### Expected behaviour

Tests that passed before, continue to pass on the same commit.

#### Actual behaviour

start/stop tests fail with:

<details>

```
=================================== FAILURES ===================================
30
_______________________________ test_start_stop ________________________________
31

32
dockerspawner_configured_app = <jupyterhub.tests.mocking.MockHub object at 0x7fb4e6a1f190>
33

34
    async def test_start_stop(dockerspawner_configured_app):
35
        app = dockerspawner_configured_app
36
        name = "has@"
37
        add_user(app.db, app, name=name)
38
        user = app.users[name]
39
        server_name = 'also-has@'
40
        spawner = user.spawners[server_name]
41
        assert isinstance(spawner, DockerSpawner)
42
        token = user.new_api_token()
43
        # start the server
44
        r = await api_request(app, "users", name, "servers", server_name, method="post")
45
        pending = r.status_code == 202
46
        while pending:
47
            # request again
48
            r = await api_request(app, "users", name)
49
            user_info = r.json()
50
            pending = user_info["servers"][server_name]["pending"]
51
        assert r.status_code in {201, 200}, r.text
52
    
53
        url = url_path_join(public_url(app, user), server_name, "api/status")
54
>       resp = await AsyncHTTPClient().fetch(
55
            url, headers={"Authorization": "token %s" % token}
56
        )
57
E       tornado.simple_httpclient.HTTPTimeoutError: Timeout during request
58

59
tests/test_dockerspawner.py:57: HTTPTimeoutError
60
----------------------------- Captured stdout call -----------------------------
61
10:23:37.452 [ConfigProxy] info: 200 GET /api/routes 
62
10:23:38.513 [ConfigProxy] info: Adding route /@/space word/user/has@/also-has@ -> http://127.0.0.1:49153
63
10:23:38.513 [ConfigProxy] info: Route added /@/space word/user/has@/also-has@ -> http://127.0.0.1:49153
64
[...]

PR #433, opened 2021-07-14
more debug info from docker when tests fail

minrk:
investigating #432

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,5 @@
 codecov
+netifaces
 notebook
 pyflakes
 pytest>=3.6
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -104,3 +104,9 @@ jobs:
       - name: Submit codecov report
         run: |
           codecov
+
+      - name: Setup tmate session
+        if: failure()
+        uses: mxschmitt/action-tmate@v3
```
minrk, line 110: tmate action makes debugging easy, thanks to @sgibson91 for the recommendation!

minrk:
Tests [passed](https://github.com/jupyterhub/dockerspawner/runs/3085517371) with 4a0c967, now to remove the debugging without breaking anything.

I'll leave the extra post-mortem report, which should hopefully prove useful. It didn't so much in this case, though.

PR #434, opened 2021-07-16
update release steps for main branch

minrk:
and update a few links

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -11,22 +11,22 @@ repository](https://github.com/jupyterhub/dockerspawner).
 
 ## Steps to make a release
 
-1. Checkout master and make sure it is up to date.
+1. Checkout main and make sure it is up to date.
 
    ```bash
    ORIGIN=${ORIGIN:-origin} # set to the canonical remote, e.g. 'upstream' if 'origin' is not the official repo
-   git checkout master
-   git fetch $ORIGIN master
-   git reset --hard $ORIGIN/master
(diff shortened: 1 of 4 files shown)


RELEASE.md
```diff
@@ -11,22 +11,22 @@ repository](https://github.com/jupyterhub/dockerspawner).
...
-1. Update [CHANGELOG.md](CHANGELOG.md) and add it to
+1. Update [docs/source/changelog.md](docs/source/changelog.md) and add it to
    the working tree.
 
    ```bash
-   git add CHANGELOG.md
+   git add docs/source/changelog.md
    ```
 
    Tip: Identifying the changes can be made easier with the help of the
```
GeorgianaElena, line 32: Since `github-activity` is now part of the `executablebooks` project (https://github.com/executablebooks/github-activity), maybe we can update the link text and URL? It's sightly unrelated to the PR, and it redirects to the right GitHub repo as it is, but I figured it would make the `RELEASE` file complete.

GeorgianaElena (approved):
Awesome :D Thanks, @minrk!
I left a comment about updating another link (not related to the branch renaming though).

(Also, using `HEAD` instead of the default branch name in GitHub links is a super nice idea)

RELEASE.md
consideRatio: Agree! I pushed a commit about that!

minrk:
Thanks, both!

PR #435, opened 2021-07-16
support cpu limit via cpu_quota / cpu_period

### LINKED ISSUE #428: Cpu Limit/Quota
This is a follow-up to #136 which was closed because not solution was found. In the mean time a working solution has been proposed in:

https://github.com/jupyterhub/dockerspawner/issues/136#issuecomment-466579224

Furthermore this was also implemented in the swarmspawner:

https://github.com/jupyterhub/dockerspawner/blob/3cdde642054887dc4c342f374a34c9cbdd843f2e/dockerspawner/swarmspawner.py#L203

minrk:
according to [the docs](https://docs.docker.com/config/containers/resource_constraints/#cpu) we need to specify cpu_quota in terms of cpu_period. The default period is 100ms

closes #428

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -29,6 +29,7 @@ from traitlets import Bool
 from traitlets import CaselessStrEnum
 from traitlets import default
 from traitlets import Dict
+from traitlets import Float
 from traitlets import Int
 from traitlets import List
 from traitlets import observe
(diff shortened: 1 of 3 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -1101,11 +1129,16 @@ async def create_object(self):
...
 
         if getattr(self, "mem_limit", None) is not None:
-            # If jupyterhub version > 0.7, mem_limit is a traitlet that can
-            # be directly configured. If so, use it to set mem_limit.
-            # this will still be overriden by extra_host_config
             host_config["mem_limit"] = self.mem_limit
 
+        if getattr(self, "cpu_limit", None) is not None:
+            # docker cpu limit is in cpu seconds per cpu period
+            # cpu_period unit is in microseconds, default is 100ms
```
consideRatio, line 1136: I didn't understand this so clearly, but not sure how to rephrase it.

consideRatio (approved):
Thanks @minrk! This LGTM, thanks for the code-review session!

PR #443, opened 2021-11-20
Test jupyterhub2

manics:
Needs a 2.0.0rc4 tag in https://hub.docker.com/r/jupyterhub/singleuser/tags to (hopefully) pass

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -9,6 +9,7 @@ import netifaces
 import pytest
 from docker import from_env as docker_from_env
 from docker.errors import APIError
+from jupyterhub import version_info as jh_version_info
 from jupyterhub.tests.conftest import app  # noqa: F401
 from jupyterhub.tests.conftest import event_loop  # noqa: F401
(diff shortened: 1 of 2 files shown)


minrk:
Looks like the prerelease tags are being rejected by our docker image-buildiung action: https://github.com/jupyterhub/jupyterhub/actions/runs/1476871178

minrk:
Ah, no, I'm wrong. jupyterhub/singleuser never got migrated to GHA from docker cloud builds!

tests/conftest.py
```diff
@@ -69,6 +70,10 @@ def named_servers(app):
 def dockerspawner_configured_app(app, named_servers):
     """Configure JupyterHub to use DockerSpawner"""
     app.config.DockerSpawner.prefix = "dockerspawner-test"
+    # If not a standard release e.g. (2, 0, 0, 'rc4', '') use full tag
+    if len(jh_version_info) > 3 and jh_version_info[3]:
```
minrk, line 74: I think this can be simpler to use just `jupyterhub.__version__`, which will always be the right string.

Issue #360, opened 2020-03-20
Any interest in PodmanSpawner ?


sudhaker:
Podman has great potential, especially with the daemonless & rootless support.

Do we have any plan to support Podman with this module? Or possibly with a separate dedicated module.

Cheers,
Sudhaker

manics:
There are no plans at the moment. There are a few ways to interface with Podman:
- Command line `podman` executable
- Varlink protocol https://github.com/containers/python-podman/blob/1.6.0/README.md
- New Docker-compatible API https://podman.io/blogs/2020/01/17/podman-new-api.html

The Docker-compatible API is still in development, but if it's similar enough it's probably reasonable to include support in this spawner. If it turns out to be quite different, or you want to use one of the other interfaces, a separate spawner makes more sense.

Though I've just noticed a post on https://github.com/containers/python-podman/issues/16#issuecomment-605439792 from @gatoniel. It looks like this is the repo https://github.com/gatoniel/podmanspawner

gatoniel:
Hi there,

I already saw this post and wanted to answer, when I could present a minimal working example. But I am not yet there. Right now I am facing the problem, that podman containers can not easily send queries to the host, which is needed for the API calls. So I will use the `--net host` option. Another option would be, that the JupyterHub itself is also run in a container and they all share the same Pod. But this is a big restriction...

Sadly, my VPN is down now and I cant connect to the server to test stuff... Hopefully I have a working Spawner tomorrow.

gatoniel:
I got it working, BUT there are some issues to be resolved. Next, I will look deeper into the new API and try to use it, based on your usage of the docker API in dockerspawner.

gatoniel:
Hi, I needed to work out some PAM issues with jupyterhub, since Podman seems to rely on the existence of /usr/run/$UID, which can easily be created with pam_systemd via pam_open_session. Now the podmanspawner works somewhat stable.

manics:
That's great! I've been meaning to give it a go but unfortunately I haven't had time so far.

minrk:
@manics @gatoniel what do you folks think now after some more experience? On a cursory read, it seems like maybe there's not that much overlap, and a separate PodmanSpawner is the right way to go.

gatoniel:
Hi,

[...]

PR #444, opened 2021-11-20
Add Podman to docs

manics:
DockerSpawner more or less works with Podman, to the point where most of the DockerSpawner tests pass! Setting up Podman as a daemon (required for compatibility with Docker) has also gotten a lot easier so now it's just a single command.

Closes https://github.com/jupyterhub/dockerspawner/issues/360

### DIFF
--- a/docs/source/spawner-types.md
+++ b/docs/source/spawner-types.md
@@ -147,3 +147,19 @@ c.DockerSpawner.host_ip = "0.0.0.0"
 This will configure DockerSpawner and SystemUserSpawner to get
 the container IP address and port number using the `docker port`
 command.
+
+## Using Podman
+
+Podman is an alternative to Docker for running containers, and supports running containers without root access.
+It is not supported by JupyterHub, but it can be used with DockerSpawner by [running a podman service](https://docs.podman.io/en/latest/markdown/podman-system-service.1.html):
+
+```sh
+podman system service --time=0 &
+export DOCKER_HOST=unix://$XDG_RUNTIME_DIR/podman/podman.sock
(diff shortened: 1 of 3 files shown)


docs/source/spawner-types.md
```diff
@@ -147,3 +147,19 @@ c.DockerSpawner.host_ip = "0.0.0.0"
 This will configure DockerSpawner and SystemUserSpawner to get
 the container IP address and port number using the `docker port`
 command.
+
+## Using Podman
+
+Podman is an alternative to Docker for running containers, and supports running containers without root access.
+It is not supported by JupyterHub, but it can be used with DockerSpawner by [running a podman service](https://docs.podman.io/en/latest/markdown/podman-system-service.1.html):
```
manics, line 154: Wasn't sure what to say here- not supported, unofficially supported, partially supported?

minrk: maybe "not officially supported"? I think "not supported" is accurate, but usually gives the impression of "doesn't work," whereas "not officially supported" is perhaps more likely to give the right impression?

Or if we do want to support it eventually, you could say "support for podman is experimental" to get at the fact that we might not be ready yet, but enthusiasts are welcome to try it out and provide feedback about what works and doesn't.

minrk (approved):

docs/source/spawner-types.md
minrk: Feel free to pick language that feels right to you and go ahead and merge.

manics: Thanks, I went with ""not officially supported"

PR #451, opened 2022-03-02
implemented pids_limit to host_config

SpaceCafe:
I implemented Docker's `pids_limit` to prevent fork bombs on JupyterHub instances.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -1114,6 +1114,28 @@ class DockerSpawner(Spawner):
         """cast cpu_limit to a float if it's callable"""
         return self._eval_if_callable(proposal.value)
 
+    pids_limit = Union(
+        [Callable(), Int(allow_none=True)],
+        help="""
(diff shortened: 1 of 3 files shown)


minrk:
Rather than duplicating every host_config field to DockerSpawner, I'm inclined to rely on `extra_host_config` for this. Does that work?

```python
c.DockerSpawner.extra_host_config = {"pids_limit": 10}
```

SpaceCafe:
> Rather than duplicating every host_config field to DockerSpawner, I'm inclined to rely on `extra_host_config` for this. Does that work?
> 
> ```python
> c.DockerSpawner.extra_host_config = {"pids_limit": 10}
> ```

Yes, that works. Thank you, I must have overlooked the option when reading the documentation 😔.

minrk:
Wonderful, thanks! Happy to add options if they are common enough and/or require substantial logic (e.g. volumes, networks), but I think where passthrough works well enough, it's the way to go.

PR #464, opened 2022-08-27
Fix imagename not to include letter ':'

yamaton:
`imagename` is currently set to like `"binder-examples-conda:HEAD"`, and a colon `:` is problematic when used as a part of a docker volume name.

```python
c.DockerSpawner.volumes = { 'jupyterhub-user-{username}--{imagename}': notebook_dir }
```

This PR sets `imagename` to be `"binder-examples-conda-HEAD"`, instead.


minrk:
Thanks!

PR #466, opened 2022-10-14
update object_name with current image

floriandeboissieu:
When several images allowed in dockerspawner and template_name is containing {imagename}, it would be expected that the selected image name is used within the object_name. I didn't found a way to do it as object_name is using default image name if not specified.
This fix updates object_name at container start after check_allowed image. It is working at least for the above use case, don't know if it is the best way to do it and if it has other consequences elsewhere. Also, I set the method _object_name_for_current_image as an async method but I am not very familiar with that method type, so here also I am not sure it is needed.

Hopping it can help, many thanks for that package anyway.


minrk:
Thanks for the PR! Sorry for the long wait.

PR #467, opened 2022-10-16
test callable allowed_images and host_ip

Sheila-nk:
- added new parameters to test allowed images passed as callables and lists. 
- mocked docker_host os environment to test that _default_host_ip correctly parses the url if the environment variable exists.

### DIFF
--- a/tests/test_dockerspawner.py
+++ b/tests/test_dockerspawner.py
@@ -18,6 +18,7 @@ from dockerspawner import DockerSpawner
 # Mark all tests in this file as asyncio
 pytestmark = pytest.mark.asyncio
 
+
 def test_name_collision(dockerspawner_configured_app):
     app = dockerspawner_configured_app
     has_hyphen = "user--foo"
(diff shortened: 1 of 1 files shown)


Sheila-nk:
The tests  fail when the callable returns a list. I am looking for feasible solutions but I would appreciate any assistance or recommendations.

Locally I hit this assertion error on Python 3.9.15:
```
>       assert resp.effective_url == url
E       AssertionError: assert 'http://127.0...Fapi%2Fstatus' == 'http://127.0...er/api/status'
E         - http://127.0.0.1:49899/@/space%20word/user/checker/api/status
E         + http://127.0.0.1:49899/@/space%20word/hub/login?next=%2F%40%2Fspace%2520word%2Fhub%2Fuser%2Fchecker%2Fapi%2Fstatus
tests/test_dockerspawner.py:133: AssertionError
```

On Github Actions the error is different though: [link to full error](https://github.com/jupyterhub/dockerspawner/actions/runs/3258873225/jobs/5351284452#:~:text=%3E%20%20%20%20%20%20%20resp%20%3D%20await%20AsyncHTTPClient,py%3A135%3A%20HTTPClientError)

```
>       resp = await AsyncHTTPClient().fetch(
            url, headers={"Authorization": "token %s" % token}
        )
E       tornado.httpclient.HTTPClientError: HTTP 500: Internal Server Error
tests/test_dockerspawner.py:135: HTTPClientError
```

tests/test_dockerspawner.py
```diff
@@ -82,27 +82,48 @@ async def test_start_stop(dockerspawner_configured_app, remove):
...
 
 
-@pytest.mark.parametrize("image", ["1.0", "1.1.0", "nomatch"])
-async def test_allowed_image(dockerspawner_configured_app, image):
+def allowed_images_callable(*_):
+    return ["jupyterhub/singleuser:1.0", "jupyterhub/singleuser:1.1"]
+
+
+@pytest.mark.parametrize(
+    "image, expected",
```
minrk, line 90: Let's call the variable what it sets, to make it clearer what it is:

- `allowed_images` is the allowed_images configuration
- `image` is the requested image

```suggestion
    "allowed_images, image",
```

tests/test_dockerspawner.py
```diff
@@ -82,27 +82,48 @@ async def test_start_stop(dockerspawner_configured_app, remove):
...
+        method="post",
+        data=json.dumps({"image": expected}),
     )
-    if image not in user.spawner.allowed_images:
-        with pytest.raises(Exception):
-            r.raise_for_status()
-        return
+
+    if not callable(image):
+        if expected not in user.spawner.allowed_images:
```
minrk, line 123: Instead of skipping this check, let's compute the allowed_images dict if it's callable:

```suggestion
    if image not in user.spawner._get_allowed_images():
```

(`_get_allowed_images()` is always correct whether `allowed_images` is callable or not)


This skip is why the test is failing - if `image` is not in `allowed_images`, the request fails and the test stops here (which is the case for your callable test - it would be good to have both an allowed and rejected case for the callable). But since the short-circuit is skipped if callable config is used, the right branch is not taken.

minrk (commented):
Thank you for the PR! I left some suggestions about why I think the changes have caused the test to fail.

Sheila-nk:
@minrk I implemented the recommended changes and the bug is fixed and tests are passing as expected.
I appreciate the feedback.

tests/test_dockerspawner.py
Sheila-nk: Great insight! Bug is fixed!

minrk:
Great job, thank you!

PR #468, opened 2022-10-19
test options_form and escape

Sheila-nk:
- set up allowed_images dictionary to test that image options are correctly displayed to user
- set up an image options formdata to test accurate conversion to user options
- added parameters of escape types to test validation of escape
- created DockerSpawner instance to test _legacy_escape on a container name

Reference: https://github.com/jupyterhub/outreachy/issues/39


minrk:
Great, thank you!

PR #470, opened 2022-10-31
catch TypeError in teardown for podman

minrk:
d.services.list throws TypeError instead of APIError with podman

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -118,8 +118,9 @@ def docker():
                 c.remove()
         try:
             services = d.services.list()
-        except APIError:
+        except (APIError, TypeError):
             # e.g. services not available
+            # podman gives TypeError
             return
         else:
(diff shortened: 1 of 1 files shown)


minrk:
doesn't get tests passing, but gets to another, real error. @manics any idea on these podman failures?

manics:
I can't see anything obvious, and the version of [Podman in the GH environment is unchanged](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2004-Readme.md). I'll do some investigation.

If this blocks merging other PRs or a release you could set `continue-on-error: true` on the Podman matrix job, and create an issue so I don't get forgotten?

consideRatio:
Closing and included this commit in #485, but the podman failure remains now with `continue-on-error: true`, and is tracked in #484.

PR #473, opened 2022-12-29
added a config option for easy gpu sharing to DockerSpawner

indirected:
I added a config option: `gpu_ids` to make sharing GPUs with containers a little bit easier and tried to keep the style close to `cpu_limit` and `mem_limit`.
Also, How it works is documented with the help parameter.

Edit: I'm kind of new to the open-source community, sorry if something is missing. comments are appreciated!

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,33 +1,5 @@
 # DockerSpawner
 
-## Fork README
-This fork adds a config option to easily share specific GPUs with containers spawned by the DockerSpawner. I have created a pull request but until it is merged, you can install the package by:
-```bash
-pip install git+https://github.com/indirected/dockerspawner.git
-```
-### Here is how you use the config option:
-```python
-c.DockerSpawner.gpu_ids = None # Will not share GPUs with containers
-c.DockerSpawner.gpu_ids = 'all' # Shares all the GPUs with containers
-c.DockerSpawner.gpu_ids = '0' # Shares GPU of ID 0 with containers
-c.DockerSpawner.gpu_ids = '0,1,2' # Shares GPUs of IDs 0, 1 and 2 with containers
-```
-Alternatively, you can pass a callable that takes the spawner as
-the only argument and returns one of the above acceptable values:
-```python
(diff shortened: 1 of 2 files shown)


minrk:
Thanks for the PR and sorry for the delays! Trying to get this repo back on track. I think in general, we do _not_ want to add new DockerSpawner-specific ways to express generic docker options that already work with the pass-through `extra_host_config`. Since I think requesting GPUs already works there, we shouldn't add gpu-specific options and instead let folks write the DeviceRequests directly. That will more appropriately direct people to the docker-py docs, rather than our own, which a custom option would do. The bit I see missing is that `extra_host_config` can't be a callable, unlike the new option here. I think the fix here should be to allow `extra_host_config` itself to be a callable, like most other options that may vary per-user. What do you think?

indirected:
Hi. Sorry for the late response. I am having a bit of academic pressure on my back right now. 
I think your argument is valid and reasonable. The only thing that I am thinking is both cpu and memory limits are also docker arguments. That is the reason I wanted to add the gpu as an additional separate option. Since it is almost a general purpose device these days and people who run jupyter servers (very much with jupyter hub and docker spawner), just like me, would appreciate a handy option to specify gpu directly as an option of docker spawner, just like memory and cpu.

One more thing is, at least at the time that I needed to add gpu, the device request part of the docker-py docs, was not complete and not helpful at all.

Again, I understand your argument and believe it is valid. I am just trying to bring ideas that how this option would benefit some people. Thanks!

minrk:
Thanks for your input! What distinguishes cpu and memory limits is that they are standardized by _jupyterhub_, so all Spawners have the same configuration for implementing a consistent resource limit. There is not an equivalent for GPUs.

If it is possible with the passthroughs we have, perhaps the right thing to add is an example showing how to do it with extra_host_config.

indirected:
Hmmm. Yes I think it is best to keep it as is but include an example. Thank you!

PR #476, opened 2023-02-13
Configure test suite to use jupyterhub pytest plugin

Sheila-nk:
- installed the JupyterHub pytest plugin through `pip`
- set `DockerSpawner` as the spawner class to be used by the hub
- setup `DockerSpawner` specific configurations
- imported the `app_for_spawners` fixture to be used in place of the JupyterHub `app` fixture


Sheila-nk:
Ping @GeorgianaElena I am getting the same errors here as well. 
Trying to go through the logs and all the errors take this form:
```python
sqlalchemy.exc.ArgumentError: Column expression, FROM clause, or other columns clause element expected, got [1]. Did you mean to say select(1)?
```
This seems to happen at the point the app is trying to connect to the database. 
Maybe you can provide more insight on this.

GeorgianaElena (commented):
Hmm, looks like the tests error because of https://github.com/jupyterhub/jupyterhub/issues/4312

Due to this, I believe the next steps are to:

- [ ] Open an issue in `dockerspawner` asking for feedback from the community about the best way to fix the sqalchemy issue, with the following context:
    - `jupyterhub` backported the `sqalchemy` compatibility fix to jupyterhub>=3.0
    - but it [didn't pin `sqalchemy` to < 2.0.0 for jupyterhub<3.0](https://github.com/jupyterhub/jupyterhub/blob/220eb87bce6f9c668ffdae4978245211ca5c9a04/requirements.txt#L14)
    - in dockerspawner we can go around this by either:
       - dropping support for jupyterhub < 3.0
       - pin `sqalchemy` to < 2.0.0 for jupyterhub<3.0 in either jupyterhub or dockerspawner

- [ ] 2. To not block this work on the first task ⬆️ add a new test strategy to the matrix to test against `jupyterhub` 3.1.1  that has fixed the `sqlalchemy` compatibility issue.
https://github.com/jupyterhub/dockerspawner/blob/8503af69161a3a543cc613f93ce7951ad30a1912/.github/workflows/test.yml#L19-L20
and only care about that tests while trying to integrate the plugin into dockerspawner.

GeorgianaElena:
> 2. To not block this work on the first task ⬆️ add a new test strategy to the matrix to test against jupyterhub 3.1.1 that has fixed the sqlalchemy compatibility issue.

@Sheila-nk, for point 2, I suggest opening a new PR, separate than this one since that is beneficial outside of this context too.

> I think it's reasonable to only support supported JupyterHub versions, it's always possible for someone to install an older version if they're still running an older JupyterHub version.

@manics, I agree 👍🏼 I suggested @Sheila-nk to propose this in https://github.com/jupyterhub/dockerspawner/issues/477 rather than have that be a bug report

GeorgianaElena:
@Sheila-nk, before analyzing this in more detail, I believe first thing should be to remove all imports from `jupyterhub's conftest` that are no longer relevant (`event_loop, ssl_tmpdir, app`, etc) to make sure things don't get mingled again and end up with different event loops and fixture mismatch.

Prior, `pytest` needed all the fixtures that the `app` fixture depended on to also be imported, event though they weren't all explicitly used in the `dockerspawner` tests. This is because an import translates into "the fixture is defined in this file" for pytest. Hence, if `app` was "defined" in dockerspaner conftest, all the other fixtures that app depended on, also needed to be imported in order for pytest to be able to find them.

But this is not the case anymore with pytest plugins, so we can safely get rid of those imports.

The "Fixture availabilty` section of the pytest docs provides **great** info and visual of this fixture execution and load order, or overrinding and even how plugin come into picture  https://docs.pytest.org/en/latest/reference/fixtures.html#fixture-availability

Sheila-nk:
> @Sheila-nk, before analyzing this in more detail, I believe first thing should be to remove all imports from `jupyterhub's conftest` that are no longer relevant (`event_loop, ssl_tmpdir, app`, etc) to make sure things don't get mingled again and end up with different event loops and fixture mismatch.

😄 I am so annoyed! I spent hours trying to find a solution to the event loop closing and all I needed to do was remove the fixtures we don't need anymore.

Thank you, @GeorgianaElena.  The tests using `JupyterHub` v3.1.1 are all passing. 🚀 
I will now pin `sqlalchemy` to 1.4 on this draft PR to ensure the plugin works for all JupyterHub versions.

Sheila-nk:
@minrk I think we might need to install `pytest-jupyterhub` from GitHub so the changes from the merged PR https://github.com/jupyterhub/pytest-jupyterhub/pull/62 can reflect here.

Issue #478, opened 2023-02-21
Discussion: SQLAlchemy2 compatibility issue


Sheila-nk:
`JupyterHub` backported the `sqalchemy` compatibility fix to JupyterHub>=3.0 but didn't restrict the `sqlalchemy` version to < 2.0 for JupyterHub<3.0. Tests on GitHub Actions are failing due to this. More info: https://github.com/jupyterhub/jupyterhub/issues/4312

Solutions suggested by @GeorgianaElena include:
- dropping support for older JupyterHub versions (see also @manics comment in https://github.com/jupyterhub/dockerspawner/pull/476#issuecomment-1436708669) or
- pinning sqalchemy in older JupyterHub's versions, or in dockerspawner

Additional input is welcome.

consideRatio:
In #485 I made us install sqlalchemy specifically when testing with jupyterhub 1 and 2, but I think we should drop support for older jupyterhub versions as well to help us maintain things a bit easier long term.

I suggest dropping support for jupyterhub 1 at least, and preferably also more, to protect us from handling issues in due time.

Here are some options:

- jupyterhub>=1.0.0 (current requirement)
- jupyterhub>=2.3.1
- jupyterhub>=3.5.1
- jupyterhub>=4.0.0

I suggest we go for `python>=3.8` and `jupyterhub>=4.0.0` to minimize long term complexity of maintaing this repo, thinking that its not so important that users without modern python or jupyterhub also get a modern version of this spawner.

consideRatio:
@minrk @manics what do you think about dropping support for older jupyterhub versions and or python versions?

I think it would be good to drop support for python 3.7 and jupyterhub 1 to enable pytest-jupyterhub to work against this repo, which requires py38+ and jupyterhub 2+ at least.

manics:
I think it's fine, especially if previous releases of DockerSpawner continue to work with older versions of JupyterHub.

minrk:
I very strongly disagree with dropping support for super recent versions, especially of hard-to-upgrade packages like JupyterHub, as I think it puts the convenience of maintainers too far above users. I think this is a temptation we should work very hard against. For example, we are still _currently_ recommending that users install JupyterHub 1.5, via the-littlest-jupyterhub, as we haven't yet made a release with the 4.0 bump. Upgrading JupyterHub is a big deal, and forcing users to upgrade JupyterHub should be considered a big cost we are passing on to users.

[...]

PR #479, opened 2023-02-22
Add test strategy for JupyterHub v3.1.1

Sheila-nk:
- removed matrix config for Python 3.6 since it reached [end of life](https://endoflife.date/python)
- added matrix config for Python 3.9 using JupyterHub 3.1.1
- added matrix config for Python 3.11 using JupyterHub 3.1.1

Reference #476

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,8 +20,6 @@ jobs:
         include:
           - python: 3.8
             test: internal-ssl
-          - python: 3.6
-            jupyterhub: "1.0"
           - python: 3.7
             jupyterhub: 1.5
           - python: 3.8
@@ -31,6 +29,8 @@ jobs:
           - python: "3.10"
             jupyterhub: "2.1"
             test: podman
+          - python: 3.11
+            jupyterhub: 3.1.1
 
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,6 +29,8 @@ jobs:
           - python: "3.10"
             jupyterhub: "2.1"
             test: podman
+          - python: 3.9
+            jupyterhub: 3.1.1
           - python: 3.11
             jupyterhub: 3.1.1
 
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,4 +5,3 @@ pyflakes
 pytest>=3.6
 pytest-asyncio
 pytest-cov
-sqlalchemy==1.*
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -5,3 +5,4 @@ pyflakes
 pytest>=3.6
 pytest-asyncio
 pytest-cov
+sqlalchemy==1.*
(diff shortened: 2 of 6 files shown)


Sheila-nk:
Ping @GeorgianaElena the tests are failing with these errors: 

`AttributeError: 'coroutine' object has no attribute 'run_sync'` and
`RuntimeWarning: coroutine 'io_loop' was never awaited`


Seems like they are still referencing the old `app` fixture from `JupyterHub`.

Sheila-nk:
io_loop error

Sheila-nk:
@GeorgianaElena Could the problem be the cached workflow dependencies?

GeorgianaElena:
> Seems like they are still referencing the old app fixture from JupyterHub.

Yes, this is because the removal of the `io_loop` from tests was not yet released. So, the [latest `jupyterhub` version 3.1.1](https://github.com/jupyterhub/jupyterhub/releases/tag/3.1.1) doesn't have the changes in https://github.com/jupyterhub/jupyterhub/pull/4332.

> AttributeError: 'coroutine' object has no attribute 'run_sync' and
RuntimeWarning: coroutine 'io_loop' was never awaited

Hmm, I believe these errors come from somewhere else and I suspect the breaking change introduced in pytest-asyncio  0.19.0 https://pytest-asyncio.readthedocs.io/en/latest/reference/changelog.html#id10 which is defaulting the `asyncio_mode` to [strict](https://pytest-asyncio.readthedocs.io/en/latest/concepts.html#strict-mode). This, together with unreliabily adding the asyncio marker to tests and fixture might have caused the use of multiple events loop.

So, my suggestion would be to explicitly add the `asyncio_mode=auto` and remove any explicit asyncio test mark and see if this fixes it.

Sheila-nk:
@GeorgianaElena the test strategies for `JupyterHub` v3.1.1 are passing as expected. There needs to be a fix for the `SQLAlchemy` compatibility issue for the other `JupyterHub` versions. I don't know whether I should focus on working on the compatibility issue: https://github.com/jupyterhub/dockerspawner/issues/478 or wait for this PR to be merged regardless so we can continue with testing the plugin.

minrk:
@Sheila-nk for sqlalchemy, you should be able to add `slqalchemy==1.*` to your test environment requirements.

Sheila-nk:
> @Sheila-nk for sqlalchemy, you should be able to add `slqalchemy==1.*` to your test environment requirements.

This worked. Thank you!

Sheila-nk:
@minrk with restricting `sqlalchemy` to v1 in `dev-requirements.txt`, is that a complete solution to the compatibility issue, or is there more to be done?

GeorgianaElena:
@Sheila-nk, I'd advocate not pinning `sqlalchemy` in `dev-requirements` in this PR.

I believe what @minrk suggested is to do experiment with thi pinning in https://github.com/jupyterhub/dockerspawner/pull/476 while integrating the pytest plugin to make sure it works. @minrk, can you please provide more details about this one?

@Sheila-nk, if you drop the commit pinning sqlalchemy, I'll be comfortable merging this PR. Additionally, I believe discussions about pinning or not pinning `sqlalchemy` to fix the tests, should happen in https://github.com/jupyterhub/dockerspawner/issues/478 for a more persistent context.

Issue #484, opened 2023-05-31
ci: test system failing


consideRatio:
Tests were [passing Aug 2 2022](https://github.com/jupyterhub/dockerspawner/actions/runs/2778505766), but were [failing Aug 9 2022](https://github.com/jupyterhub/dockerspawner/actions/runs/2821509933), specifically the podman job started failing.

consideRatio:
Working to fix these in #485, including misc updates to this repo to catch up with chore maintenance.

PR #485, opened 2023-05-31
Refresh test workflow and associated config, accept podmain test failure for now

consideRatio:
This PR updates the very old test workflow and associated configuration, allowing all tests besides one tracked by #484 to pass successfully. Pinning sqlalchemy to version 1 was a key thing done but there was several smaller things that needed to be done. I've made a few comments about those changes below.


minrk:
Last attempt at the podman tests was #470

consideRatio:
It seems it comes down to this:

>     docker.errors.APIError: 500 Server Error for http+docker://localhost/v1.40/containers/65b02b7e1a0aa53a877e961e1f65b3af4be0c89c20fe4d9bc4d2bc619f816002/start: Internal Server Error ("the requested cgroup controller `cpu` is not available: OCI runtime error")

.github/workflows/test.yaml
```diff
@@ -0,0 +1,132 @@
...
+            test-variation: podman
+
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "18"
```
consideRatio, line 56: I updated to run CHP with node 18 from 14.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+      - name: setup docker swarm
+        run: docker swarm init
+
+      - name: Install Python dependencies
+        run: |
+          pip install ${{ matrix.pip-install-spec }}
+          pip install -e "." -r dev-requirements.txt
```
consideRatio, line 64: I installed dependencies a bit more generally, allowing me to pin sqlalchemy without too much hard to debug bash logic.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,132 @@
...
+
+      - name: List Python dependencies
+        run: |
+          pip freeze
```
consideRatio, line 68: I added a separate step to list dependencies, allowing for easier inspection of differences between test runs.

I also removed a pip cache step because it makes understanding failures associated with new releases of python packages quite a bit harder to track down.

.github/workflows/test.yaml
```diff
@@ -0,0 +1,132 @@
...
+          #
+          #        If this is resolved, also remove the continue-on-error
+          #        section above.
+          #
+          cat "/sys/fs/cgroup/user.slice/user-$(id -u).slice/user@$(id -u).service/cgroup.controllers"
+
+          pytest tests/test_dockerspawner.py --cov=dockerspawner
+
+      # GitHub action reference: https://github.com/codecov/codecov-action
+      - uses: codecov/codecov-action@v3
```
consideRatio, line 132: I updated this project to use the non-deprecated github action.

minrk (approved):

Issue #453, opened 2022-04-05
Volume owned by root


keshav-1835i:
### Bug description
Below are lines from my `jupyter_config.py`
`
c.DockerSpawner.volumes = { 
'jupyterhub-user-{username}':'/home/jovyan/work',
'jupyterhub-shared': {"bind": '/home/jovyan/shared', "mode": "ro"}
}
`
`/home/jovyan/work` is owned by jovyan as expected
#### Expected behaviour
`/home/jovyan/shared` is owned by jovyan 

#### Actual behaviour
`/home/jovyan/shared` is owned by root

### How to reproduce

Spin up a user container with above config

### Your personal set up

 - OS:
 - ubuntu 20.04
 - docker latest
 - latest jupyterhub and dockerspawner

Any help would be much appreciated

minrk:
Maybe this has changed, but I think it's up to the docker volume provider to set initial permissions. We don't expose volume creation options in DockerSpawner, yet.

I believe the `work` ownership is set by the fact that the directory exists in the image, while `shared` doesn't so it uses the default of `root:root`.

If the empty `shared` mount point exists in your image with the right ownership, it may work correctly.

Alternately, you can manage creation of the `shared` volume and its ownership/permissions outside jupyterhub. I assume you are doing this already, since if it's mounted read-only there must be something else with write access to put things there.

Since it's read-only, is there an issue with it being owned by root?

keshav-1835i:
Thank you so much @minrk for quick response. I was lazy and 'ro' was copied from other issue. I tried 'rw' too. 

I will follow your lead and see if i can have the shared directory present with right perms

matercomus:
I have the same issue. The only difference is that I have the shared folder created inside of work so it shows up in explorer by default.
`c.DockerSpawner.volumes = { 
'jupyterhub-user-{username}':'/home/jovyan/work',
'jupyterhub-shared': {"bind": '/home/jovyan/work/shared', "mode": "rw"}
}`
But even with `"mode": "rw"` I still cannot access it from the spun-up container. Any help would be appreciated!

matthewwiese:
In the interest of visibility, I also encounter this problem. The related part of my config:

```python
notebook_dir = "/home/jovyan/work"
c.DockerSpawner.notebook_dir = notebook_dir
c.DockerSpawner.volumes = { "/mnt/volume_nyc3_01/hub_userdata/{username}": notebook_dir }
```

[...]

Issue #154, opened 2017-04-21
Granting users admin permissions on their container


kjschiroo:
How do you give users permission to install packages on their server?

Specifically I am trying to connect to mssql, as part of this I need to install `freetds`

Here are the two paths I go down:
```
$ apt-get install freetds
E: Could not open lock file /var/lib/dpkg/lock - open (13: Permission denied)
E: Unable to lock the administration directory (/var/lib/dpkg/), are you root?
```
So I try it with `sudo`
```
$ sudo apt-get install freetds

We trust you have received the usual lecture from the local System
Administrator. It usually boils down to these three things:

    #1) Respect the privacy of others.
    #2) Think before you type.
    #3) With great power comes great responsibility.

[sudo] password for jovyan:
```
and I don't know the password for jovyan.

yuanzhou:
I have the same issue here. I tried to find the `start-singleuser.sh` but can't find it.

minrk:
If you are using docker-stacks and set the environment `GRANT_SUDO=1`, you should be able to allow this:

```python
c.DockerSpawner.environment = {
    'GRANT_SUDO': '1',
}
```

@yuanzhou that's unlikely to be related. `start-singleuser.sh` is in `/usr/local/bin/` in the existing docker-stacks

tovmeod:
I'm using docker-stacks as base image, but GRANT_SUDO doesn't seems to work:

c.DockerSpawner.container_image = 'jupyter/datascience-notebook'
c.DockerSpawner.environment = {
    'GRANT_SUDO': '1',
}

but it still asks for a password when running !sudo ls

Running !ls /etc/sudoers.d just returns README, no /etc/sudoers.d/notebook

minrk:
@parente I think you know the sudo stuff on docker-stacks better than I do.

minrk:
@tovmeod I was just looking into this, and I think there's more work needed to get the root-level permissions using docker-stacks, which are required to grant sudo permissions. Try this:

```python
c.DockerSpawner.container_image = 'jupyter/datascience-notebook'
# must start container as root in order to grant sudo permissions
c.DockerSpawner.extra_create_kwargs = {'user': 'root'}
c.DockerSpawner.environment = {
  'GRANT_SUDO': '1',
  'UID': '0', # workaround https://github.com/jupyter/docker-stacks/pull/420
}
```

kjschiroo:
The way that I got around this was creating a new container image with a docker file like this: 
```
[...]

PR #492, opened 2023-07-28
Add initial_user traitlet to change user that container is started with

matthewwiese:
This PR adds the `initial_user` traitlet, which may be used to change the user that the spawned container is started with.

Per my [comment here](https://github.com/jupyterhub/dockerspawner/issues/453#issuecomment-1640502595) I was encountering a mounted directory with `root` permissions that meant the default `jovyan` user had no access.

In order to follow the [suggestion here](https://github.com/jupyter/docker-stacks/issues/1003#issuecomment-583768897), I needed DockerSpawner to start the container as `root` so that the permissions may be changed. This can be done via this traitlet like so:

```python
c.DockerSpawner.initial_user = "root"
```

This PR is inspired by [this comment](https://github.com/jupyterhub/dockerspawner/issues/154#issuecomment-355477006), which I saw was not yet implemented. It is useful beyond my specific case, such as when changing [user-related configuration](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/common.html#user-related-configurations) in a Jupyter Docker Stack

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -195,6 +195,23 @@ class DockerSpawner(Spawner):
     def _ip_default(self):
         return "0.0.0.0"
 
+    initial_user = Unicode(
+        "jovyan",
+        help="""
+        The initial user with which to start the container
+
+        The [default user for Jupyter Docker stacks](https://jupyter-docker-stacks.readthedocs.io/en/latest/using/selecting.html#jupyter-docker-stacks-foundation) is 'jovyan'
+
+        You may need to set this value to 'root', which is equivalent to
+        `docker run --user root`. This is necessary when e.g. changing jovyan's
+        default UID or home directory permissions
+
+        Jupyter Docker stacks will switch back to jovyan before starting the notebook server (e.g. JupyterLab)
+        The end-user within the container *will not* be the user set here, unless you are using a modified image
+        """,
+        config=True,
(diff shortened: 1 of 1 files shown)


matthewwiese:
I have discovered that this can just be done via the existing `extra_create_kwargs`:

```python
c.DockerSpawner.extra_create_kwargs = {
    "user": "root"
}
```

I can close this PR as it's superfluous given the above knowledge.

Related: [`tests/test_dockerspawner.py::test_post_start`](https://github.com/jupyterhub/dockerspawner/blob/main/tests/test_dockerspawner.py#L207) fails for this PR because there is no `jovyan` user in the [`busybox:1.29.1` image](https://github.com/jupyterhub/dockerspawner/blob/main/tests/test_dockerspawner.py#L223).

manics:
I think `extra_create_kwargs` is a good solution, especially since if you add a user config it should also allow a numeric UID, and it should default to the containers default.

Do you think it'd be useful to add this as an example in the documentation?

matthewwiese:
> Do you think it'd be useful to add this as an example in the documentation?

Yeah, I think an example in the docs would be handy, if only to increase the visibility of `extra_create_kwargs`. When I was investigating the problem that lead to this PR, I had read the entry for `extra_create_kwargs` in the API reference but it didn't immediately dawn on me how it may be useful.

PR #493, opened 2023-08-03
Add extra_create_kwargs example, plus docs readability improvements

matthewwiese:
This PR adds an example to the API reference for `extra_create_kwargs`.

Additionally, it fixes inline code formatting by adding an additional `` ` `` enclosing each end of the text. The existing usages involve single backticks ala Markdown, whereas [double backticks are used in RST](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#inline-markup).

Continued from #492 

cc @manics


manics (approved):
Thanks!

PR #508, opened 2023-11-20
Changelog for 13

minrk:
Small major release

### DIFF
--- a/docs/source/changelog.md
+++ b/docs/source/changelog.md
@@ -8,7 +8,7 @@ command line for details.
 
 ## 13
 
-### 13.0 2023-11
+### [13.0] 2023-11-21
 
 ([full changelog](https://github.com/jupyterhub/dockerspawner/compare/12.1.0...13.0.0))
 
@@ -63,7 +63,7 @@ See [our definition of contributors](https://github-activity.readthedocs.io/en/l
 
 ## 12
 
-### [12.1] 2021-07
+### [12.1] 2021-07-22
 
 ([full changelog](https://github.com/jupyterhub/dockerspawner/compare/12.0.0...12.1.0))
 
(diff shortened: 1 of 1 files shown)


docs/source/changelog.md
```diff
@@ -6,6 +6,61 @@ command line for details.
 
 ## [Unreleased]
 
+## 13
+
+### 13.0 2023-11
+
+([full changelog](https://github.com/jupyterhub/dockerspawner/compare/12.1.0...955e64222447c22df6a708b7d63d90059303337d))
```
manics, line 13:
```suggestion
([full changelog](https://github.com/jupyterhub/dockerspawner/compare/12.1.0...13.0.0))
```

manics (commented):
Other than the changelog diff link, this LGTM.

docs/source/changelog.md
```diff
@@ -6,6 +6,61 @@ command line for details.
 
 ## [Unreleased]
 
+## 13
+
+### 13.0 2023-11
```
consideRatio, line 11: A link update in the end of the file is needed for consistency here.

I think having the exact date is very helpful as well because I often track down _why did this break now?_ and look for changes between specific dates in dependencies. This is not what we have done historically in this file though.

consideRatio (approved):
Wieee, LGTM but commented on a detail

docs/source/changelog.md
manics: Good point! It's only the 12.* releases that are missing the full date, [earlier releases have them](https://github.com/jupyterhub/dockerspawner/blob/main/docs/source/changelog.md#011)

manics: If you want to fill them in now the [tag dates](https://github.com/jupyterhub/dockerspawner/tags) are
12.1.0: 2021-07-22
12.0.0: 2021-03-26

minrk:
Thanks! Fixed links and dates

PR #509, opened 2023-11-20
Update dockerspawner.py support mount `type` args

zhangnew:
```python
class Mount(target, source, type='volume', read_only=False, consistency=None, propagation=None, no_copy=False, labels=None, driver_config=None, tmpfs_size=None, tmpfs_mode=None)
```

see https://docker-py.readthedocs.io/en/stable/api.html#docker.types.Mount


manics:
Can you explain why this is needed? `args = dict(mount)` already incorporates those arguments if they've passed, and if they're not then the Docker defaults are used.

PR #525, opened 2024-12-24
Convert options_from_form into traitlet

### LINKED ISSUE #524: DockerSpawner breaks the Spawner options_form API
### Bug description

jupyterhub.spawner.Spawner defines `options_from_form` as a Callable traitlet. `dockerspawner.DockerSpawner` instead defines `options_from_form` as a method. This prevents overriding `options_from_form` without subclassing. Is there a reason for this change?

I arrive here from investigating a related question: https://discourse.jupyter.org/t/more-love-to-spawner-options-form/31126?u=anton_akhmerov

akhmerov:
Closes #524.

I believe similar changes should also be done to other spawners; didn't check those yet.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -330,12 +330,16 @@ class DockerSpawner(Spawner):
             options=options
         )
 
-    def options_from_form(self, formdata):
-        """Turn options formdata into user_options"""
-        options = {}
-        if 'image' in formdata:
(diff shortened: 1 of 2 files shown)


akhmerov:
The test failure seems to be unrelated. I'm happy to implement similar changes to other spawners and tests, although not sure what is there to test.

minrk:
Thanks!

PR #530, opened 2025-02-12
require Python 3.9, jupyterhub 4, unpin pytest-asyncio

minrk:
increase lower bound on pytest-asyncio to 0.24

this may create a problem for the older jupyterhub tests due to imported fixtures, but let's see...

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -31,10 +31,13 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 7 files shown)


minrk:
increased required python to 3.9 (oldest Python with security support), jupyterhub to 4 (oldest jupyterhub without known vulnerabilities) because the imported jupyterhub fixtures have complex interactions with pytest-asyncio going back too many versions.

PR #535, opened 2025-04-04
Cleanup how mounts are handled

yuvipanda:
- Don't require specific keys (like 'source') to be present, as there are mount types that don't need that (like https://docs.docker.com/engine/storage/tmpfs/)
- Format all values, not just source and destination.
- Remove two layers of indirection - mount_binds was only used in one place, so let's just use it there.

### DIFF
--- a/dockerspawner/dockerspawner.py
+++ b/dockerspawner/dockerspawner.py
@@ -442,7 +442,7 @@ class DockerSpawner(Spawner):
     )
 
     mounts = List(
-        Dict,
+        Dict(),
         config=True,
         help=dedent(
             """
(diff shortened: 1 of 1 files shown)


dockerspawner/dockerspawner.py
```diff
@@ -442,6 +442,7 @@ def _default_name_template(self):
     )
 
     mounts = List(
+        Dict,
```
minrk, line 445: should be an instance instead of a type:

```suggestion
        Dict(),
```

minrk (approved):
LGTM! Is there an easy test to add for this? I don't think mounts get exercised at all, yet. If it's not simple, no worries, we can add an Issue for test coverage of mounts and still merge this.

Issue #536, opened 2025-05-13
SwarmSpawner volumes not working as expected


kreuzert:
### Bug description

When using SwarmSpawner in a DockerSwarm environment, volume mounting does not work as expected.

### How to reproduce

Create a local DockerSwarm Setup
```
docker swarm init
docker network create --driver=overlay --attachable jupyterhub-net
docker pull jupyter/base-notebook:latest
```

<details><summary>Docker Compose File</summary>
Minimal docker compose file:  

```yaml
# jupyterhub-compose.yaml
version: "3.8"

services:
  jupyterhub:
    entrypoint:
      - /bin/bash
      - -c
      - "pip install dockerspawner==13 ; jupyterhub"
    image: jupyterhub/jupyterhub:latest
    ports:
      - "8000:8000"
      - "8081:8081"
    volumes:
      - jupyterhub-data:/srv/jupyterhub
      - /var/run/docker.sock:/var/run/docker.sock

    environment:
      DOCKER_NETWORK_NAME: jupyterhub-net
      SPAWNER_CLASS: dockerspawner.SwarmSpawner
      HUB_IP: jupyterhub
    deploy:
      replicas: 1
      placement:
        constraints: [node.role == manager]
    networks:
      - jupyterhub-net

volumes:
  jupyterhub-data:

networks:
  jupyterhub-net:
    external: true
```

</details>
<br>
<details><summary>JupyterHub Configuration</summary>  

Example jupyterhub_config.py file:  
```python
# jupyterhub_config.py
c = get_config()

from dockerspawner import SwarmSpawner

c.JupyterHub.authenticator_class = "dummy"

c.JupyterHub.ip = '0.0.0.0'
c.JupyterHub.hub_connect_url = "http://jupyterhub:8081"
c.JupyterHub.hub_bind_url = "http://0.0.0.0:8081"
c.JupyterHub.log_level = 10

c.JupyterHub.spawner_class = SwarmSpawner
c.SwarmSpawner.image = 'jupyter/base-notebook:latest'
c.SwarmSpawner.network_name = 'jupyterhub-net'
c.SwarmSpawner.debug = True

### Use persistent storage example from https://jupyterhub-dockerspawner.readthedocs.io/en/latest/data-persistence.html
notebook_dir = '/home/jovyan/work'
c.SwarmSpawner.notebook_dir = notebook_dir
# Mount the real user's Docker volume on the host to the notebook user's
# notebook directory in the container
[...]

PR #537, opened 2025-05-13
add mount type for volume_binds generated mounts

kreuzert:
See #536 for more information

### DIFF
--- a/dockerspawner/swarmspawner.py
+++ b/dockerspawner/swarmspawner.py
@@ -116,6 +116,16 @@ class SwarmSpawner(DockerSpawner):
             )
         return None
 
+    volume_binds_mount_type = Unicode(
+        default_value="bind",
+        allow_None=False,
+        config=True,
+        help="""
+        Use this mount type for creating mounts of volume_binds.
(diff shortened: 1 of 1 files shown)


minrk:
I think this might be fixed by #535 which lets you pass through all mount arguments without needing any new config fields. Does that work for you?

kreuzert:
That's looking interesting. This might fix the issue for persistent user storage volumes. But I'm not sure it will solve the issue for "internal-ssl-certs" volumes. They're not configured previously by the config, so I cannot configure the required mount arguments. 

Or am I missing an option to mount the ssl volume with different arguments?

minrk:
right, that makes sense. I wonder if there's a better passthrough here, because in general I want to avoid exposing underlying options one at a time if it's avoidable, and instead let you passthrough things directly to the docker API.

Do you think there's perhaps an _equivalent_ to #535 for swarm mounts? Does it work to specify type via `volume_driver_options` and _remove_ `type` from the Mount kwargs?

Re-reading a bit, I think maybe the `type='bind'` is actually just wrong for swarm, though it works in single-node tests.

